PR #21, opened 2017-08-12
Add Travis support

### LINKED ISSUE #1: Integrate Travis-CI
Per our previous CI environment this should build with Rust stable, beta and nightly.

If possible the two releases prior to the current stable Rust also.

As part of this ticket Cargo.toml should be updated with Travis-CI badges.

### LINKED ISSUE #4: Calculate code coverage
Integrate with web tooling, such as coveralls.io, codecov.io or codeclimate.com so this information is easily accessible per PR.

PlasmaPower:
You'll need to enable Travis for the repository. I'd recommend doing that *before* this PR gets merged, so we can check that the `.travis.yml` works (note: turn off the "build without .travis.yml" option so it doesn't try to use the default ruby build on master).

Fixes #1 

From that issue:

> If possible the two releases prior to the current stable Rust also.

Unfortunately that's rather hard to do without hard coding the versions 🙁 . Maybe in a future PR (or if you really want it I guess I could try with this PR).

Edit: now also fixes #4 with coveralls integration through travis-cargo. You'll need to use enable coveralls for the repository to activate that (through their website).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: rust
+rust:
+  - stable
+  - beta
+  - nightly
+matrix:
+  allow_failures:
+    - rust: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+cache: cargo
 rust:
   - stable
   - beta
@@ -17,7 +18,7 @@ addons:
 env:
   - PATH=$HOME/.cargo/bin:$PATH
 before_script:
-  - cargo install cargo-travis
+  - cargo install cargo-travis -f
 script:
   - cargo test
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,22 @@ rust:
   - stable
   - beta
   - nightly
+# necessary for travis-cargo coveralls
+addons:
+  apt:
+    packages:
+      - libcurl4-openssl-dev
+      - libelf-dev
+      - libdw-dev
+      - binutils-dev
+env:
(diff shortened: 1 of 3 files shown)


smangelsdorf:
Closing / reopening this to try trigger a Travis CI build.

smangelsdorf:
Thanks for this, it's looking good! Everything appears to work right up until the attempt to upload to coveralls, where it appears that `kcov` isn't being invoked correctly:

```
Uploading coverage
Usage: kcov [OPTIONS] out-dir in-file [args...]
```

> > If possible the two releases prior to the current stable Rust also.
>
> Unfortunately that's rather hard to do without hard coding the versions

Let's skip it and call #1 fixed. Perhaps there will be an easier way to tackle this in the future.

PlasmaPower:
Looks like https://github.com/huonw/travis-cargo/issues/72

I guess `travis-cargo coverage` is just broken..?

PlasmaPower:
Oh, I guess `travis-cargo` isn't maintained. Switching to `cargo-travis` should work. I'd like to get caching going too, I'll figure that out next.

PlasmaPower:
Caching should be working, but I'm not sure if `cargo-travis` itself is cached. I think they'll eventually have binary releases.

bradleybeddoes:
Once this is all working we should apply the same config to https://github.com/gotham-rs/example-app.

PlasmaPower:
@bradleybeddoes I think this is working

smangelsdorf:
Looks like it's all working to me: Travis build, caching, Coveralls stats and line coverage reporting.

Thanks again.

PR #28, opened 2017-08-17
Fix the blog URL

samsymons:
It looks like the [current blog URL](https://staging.gotham.rs/blog) is not working and is pointing to a staging environment — this should point to the [regular blog URL](https://gotham.rs/blog) instead.


bradleybeddoes:
Thanks!

PR #29, opened 2017-08-18
Make src/test/mod.rs appear only in test config

target-san:
The file named uses some Unix-only features for testing. This prevents Gotham to be used on non-Unix systems, event if people don't run tests there. Adding module-wide configuration gate should prevent it from compiling in non-test configurations.


target-san:
Sorry, missed the point that test module is used in doctests. Dropping PR for now.

PR #30, opened 2017-08-19
Fix support for Windows

smangelsdorf:
Our usage of unix sockets for `TestServer` was preventing compilation on Windows. Originally, unix sockets were introduced to avoid having magic URLs in use for integration testing, essentially getting an anonymous, full-duplex pipe.

Another alternative here would be named pipes, however using TCP means we're closer to the real production code for integration tests, and we can support all platforms from a single implementation.

After seeing #29 from @target-san, and @braunse's fork, I realised this was a higher priority than I initially thought.

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -5,9 +5,7 @@
 //! [TestServer::new]: struct.TestServer.html#method.new
 
 use std::{cell, io, net, time};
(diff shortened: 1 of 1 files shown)


bradleybeddoes:
Really great to get this support in!!

developers

PR #36, opened 2017-08-25
Improve concurrency by removing CpuPool from NewHandlerService

smangelsdorf:
Since 032249ac we've been using a `CpuPool` in `NewHandlerService` for concurrency, however there are a couple of downsides to this:

* We're bound by the single-thread performance of `hyper`
* We've got `Send` and `Sync` trait bounds in a lot of places where they wouldn't be otherwise necessary

This change moves the concurrency to the top level, starting a Tokio reactor per thread (defaulting to one thread for each CPU), and removes unnecessary `Send` / `Sync` trait bounds. I'm seeing about a 25% performance boost from this approach, just based on some very unscientific benchmarking.

Concept is based on [this gist from Alex Crichton](https://gist.github.com/alexcrichton/7b97beda66d5e9b10321207cd69afbbc), but some refactoring was needed first to attempt to support Windows, and then to ultimately use a different code path when apparently unable to share the same server socket across threads in Windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ base64 = "0.4"
 rand = "0.3"
 linked-hash-map = "0.4"
 num_cpus = "1"
+crossbeam = "0.3"
 
 [dev-dependencies]
 gotham_derive = { version = "0.1.0", path = "gotham_derive" }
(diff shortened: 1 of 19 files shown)


smangelsdorf:
This also addresses (by pure coincidence) the current build failure on the `master` branch, which is due to `Future::boxed()` being deprecated. (We're now using `Box::new(..)` instead, because `Future::boxed()` required `Self: Send`)

PR #49, opened 2017-09-25
Fix blank session cookie bug

smangelsdorf:
This looks like a refactor gone wrong in b0b39047, which was missed and got merged after the initial (pre-refactor) version had already been reviewed. The problem is that an empty `Vec` with a capacity is still seen as a 0-length slice, since none of the elements are initialized.

The RNG attempts (quite successfully) to fill the 0-length slice, and the session is established with a 0-length identifier.

Originally reported in https://github.com/gotham-rs/example-app/issues/5


smangelsdorf:
Expecting doctest failures due to new deprecation warnings we've fixed in `master`

bradleybeddoes:
Noting CI error output is not "real", have also verified fix locally.

PR #51, opened 2017-09-26
Refine SessionCookieConfig implementation

bradleybeddoes:
The previous implementation was not ideal when we came to implement
configurable path values, mostly in that it was complex to update an
option that had runtime data associated with it.

The new implementation simply sets values on the SessionCookieConfig
struct directly making updates easier and future extensions simpler.

External usage should not be impacted by this change as the
existing public API was able to be kept and in the case of
`anchor_to_path` extended.

### DIFF
--- a/src/middleware/session/mod.rs
+++ b/src/middleware/session/mod.rs
@@ -1,6 +1,6 @@
 //! Defines a default session middleware supporting multiple backends
 
-use std::{io, fmt};
+use std::io;
 use std::sync::{Arc, Mutex, PoisonError};
 use std::ops::{Deref, DerefMut};
 use std::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


src/middleware/session/mod.rs
```diff
@@ -51,49 +51,73 @@ enum SessionDataState {
     Dirty,
 }
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, PartialEq, Debug)]
```
smangelsdorf: Is this `PartialEq` still necessary?

src/middleware/session/mod.rs
```diff
@@ -51,49 +51,73 @@ enum SessionDataState {
...
+    domain: Option<String>,
+}
+
+impl Default for SessionCookieConfig {
+    fn default() -> SessionCookieConfig {
+        SessionCookieConfig {
+            name: "_gotham_session".to_string(),
+            secure: true,
+            http_only: true,
+            same_site: SameSiteEnforcement::Strict,
```
smangelsdorf: The prior default was `SameSiteEnforcement::Lax`, which was chosen after careful deliberation. I propose changing it back unless there's a compelling reason for the new default.

Seems likely that this is an oversight, because the `with_strict_same_site_enforcement` function still exists but is a noop now.

src/middleware/session/mod.rs
```diff
@@ -601,7 +633,11 @@ where
     /// # ;}
     /// ```
     pub fn allow_cross_site_usage(self) -> NewSessionMiddleware<B, T> {
-        self.remove_cookie_option(CookieOption::SameSite(SameSiteEnforcement::Lax))
+        let cookie_config = SessionCookieConfig {
+            same_site: SameSiteEnforcement::Lax,
+            ..(*self.cookie_config).clone()
+        };
+        self.rebuild_new_session_middleware(cookie_config)
```
smangelsdorf: Rustdoc for this function:

> Removes the `SameSite` cookie attribute, allowing cross-site requests to include the cookie.

We currently have no mechanism to remove the `SameSite` attribute, so we'll need a `None` variant, or to wrap the field in an `Option<_>` to implement this correctly.

src/middleware/session/mod.rs
```diff
@@ -440,51 +464,55 @@ where
...
+    ///     .with_session_type::<MySessionType>()
+    ///     .anchor_to_path("/app".to_string())
+    /// # ;}
+    /// ```
+    pub fn anchor_to_path<P>(self, path: P) -> NewSessionMiddleware<B, T>
+    where
+        P: AsRef<str>,
+    {
+        let cookie_config = SessionCookieConfig {
+            name: path.as_ref().to_owned(),
```
smangelsdorf: `name` here should be `path`

src/middleware/session/mod.rs
bradleybeddoes: yes, used in tests

Issue #14, opened 2017-08-08
Migrate KitchenSink examples into integration tests


bradleybeddoes:
KitchenSink is a great example of the various pieces of Gotham functioning together but is has grown to the point where maintaining it manually is no longer feasible.

We'll get much better value by splitting this up into integration tests that can be run through CI whilst still maintaining its original purpose as a set of basic examples for working with each component of Gotham.

bradleybeddoes:
Incorrectly closed during source control system migration.

chklauser:
Hi,

I decided to have a go at #14, mostly as an exercise to get to know gotham a bit better. My development branch is over here: [migrate-kitchen-14:./tests](https://github.com/chklauser/gotham/tree/migrate-kitchen-14/tests) ([diff](https://github.com/gotham-rs/gotham/compare/master...chklauser:migrate-kitchen-14)).

Before I continue, I would like to know if this is broadly going into the direction you had imagined and maybe what your
 opinion is on some of my decisions.

## Splitting up `kitchen-sink`
As its name suggests, the `kitchen-sink` example demonstrates a number of different features together in one
application. I intend to split it up into 3 separate test crates:
* `echo_server.rs` Just static routes. Establishes a base level for the setup code that gotham needs.
* `dynamic_xxx.rs` (TODO) dynamic routes, path extractors and query extractors
* `middleware_server.rs` Just the middleware with some handler that interacts with it

Each test create can run multiple tests against the test server it sets up.

**Would you prefer a different split?**

## `TestServer` vs. `gotham::start`
The `kitchen-sink` example was a full-blown example application, running multiple listener-threads and spawning a
handler thread for each request. Initially I had a look at the `TestServer` struct, but it seems more targeted towards
unit testing (single threaded).

I decided to stick closer to `kitchen-sink` and wrote a small function that lets multiple tests (run concurrently by
cargo) share a single server implementation. One of the tests will be the first to secure a Mutex and gets to start a
the server. The other tests wait until the server accepts TCP connections.

Seemed to work reliably enough, but I'm not sure if that's what you guys are looking for. Technically, all of the
interesting stuff happens in the `build_router()` function and unless we start spawning threads in the individual test
functions, there is probably no between `TestServer` and "the real thing". Personally, I think that the integration
tests would be the place to use "real" implementations where possible.

(One TODO in my implementation: need to ensure the `build_router()` function only runs once.)

**Which approach seems more useful to you?**

## HTTP Client Library
[...]

PR #52, opened 2017-09-26
Refactor TestServer API

### LINKED ISSUE #16: Improve TestServer API
The 0.1 API for `TestServer` is a pretty rough, early version. Due to the way we're handling connections internally and allowing _everything_ to be configured, there are a lot of arbitrary variables being set by test cases when they could be defaulted and only overridden when the need arises.

This ticket is mostly a "go back and reduce the boilerplate for integration tests" ticket.

The way a test HTTP client is created and used seems like an obvious target.

smangelsdorf:
The big pieces here:

* Retain the client address in `State` - it was previously available via `Request`, but after #45 we're no longer able to access it
* Refactor `TestServer` to take a `NewHandler` (most commonly this will be a `Router`) instead of a `NewService`. Previously, the test case needed to wrap its own handler in `NewHandlerService`
* Reduce the boilerplate needed to use `TestServer`, and make the API a little more intuitive

Before:

```rust
let new_service = NewHandlerService::new(router);
let mut test_server = TestServer::new(new_service).unwrap();
let client = test_server
    .client("127.0.0.1:10000".parse().unwrap())
    .unwrap();
let uri = "http://example.com/".parse().unwrap();
let response = test_server.run_request(client.get(uri)).unwrap();
assert_eq!(response.status(), StatusCode::Ok);
assert_eq!(
    test_server.read_body(response).unwrap(),
    b"Hello, world!"
);
```

After:

```rust
let test_server = TestServer::new(router).unwrap();
let response = test_server.client().get("http://example.com/").perform().unwrap();
assert_eq!(response.status(), StatusCode::Ok);
assert_eq!(response.read_body().unwrap(), b"Hello, world!");
```

Fixes #16

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -15,6 +15,7 @@ use hyper::{Request, Response};
 use futures::{future, Future};
 
 use state::{State, set_request_id};
+use state::client_addr::put_client_addr;
 use http::request::path::RequestPathSegments;
 
 mod error;
@@ -148,9 +149,14 @@ where
     type Future = Box<Future<Item = Self::Response, Error = Self::Error>>;
 
     fn call(&self, req: Self::Request) -> Self::Future {
+        let mut state = State::new();
+
+        if let Some(addr) = req.remote_addr() {
+            put_client_addr(&mut state, addr);
+        }
+
         let (method, uri, version, headers, body) = req.deconstruct();
 
-        let mut state = State::new();
         state.put(RequestPathSegments::new(uri.path()));
         state.put(method);
         state.put(uri);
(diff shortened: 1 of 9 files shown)


smangelsdorf:
Hold this, pending outcomes of discussion in #14

bradleybeddoes:
Are we still holding pending https://github.com/gotham-rs/gotham/issues/14 ?

smangelsdorf:
Yes. Sorry, I've been waylaid by other work and this still needs to be sorted out.

smangelsdorf:
@chklauser: I've added some changes in here that you suggested in #14. If you have time to look at them, I'd appreciate any further suggestions or comments.

PR #58, opened 2017-10-16
Remove deprecated BoxFuture and boxed

king6cong:
as noted here: https://github.com/alexcrichton/futures-rs/pull/556


smangelsdorf:
Hello, thanks for noticing this. Actually, we already sorted this out in #36 by pure coincidence (see comments), which is the reason you're seeing conflicts in this PR. If you fast-forward your branch to our latest `master`, the deprecation warnings should be resolved.

king6cong:
Great

PR #59, opened 2017-10-17
Bump chrono version to 0.4.0

r-bar:
Small update to make gotham compatible with the latest version of `chrono`.

Edit: After a quick search of your other repositories I noticed gotham-rs/example-app also uses the older 0.3.0 version. I wanted to open a pull request that repo as well, however that would cause a version mismatch with chrono for gotham 0.1.0. Is there an example-app that tracks gotham master?


smangelsdorf:
Welcome. Thanks for updating this. 👍 

> Is there an example-app that tracks gotham master?

No, the idea with the example-app repo (at least for now) is that it targets the most recently _released_ version of Gotham. At the moment that means it's a bit out of step with all the changes between 0.1 and 0.2, but over time I expect the amount of API churn to slow down, and for that not to be as large a problem.

We'll revisit whether we need a version targeting `master` once we get a bit further along.

PR #61, opened 2017-10-27
Add support for catching panics from futures

### LINKED ISSUE #17: A panic inside a handler should return HTTP 500, not kill the server
Running the example app, it's easy to cause a panic with this message:

    thread '<unnamed>' panicked at 'attempt to add with overflow', src/controllers/challenge.rs:34

simply by visiting this address: http://127.0.0.1:7878/challenge/Rustacean?count=255

If a handler panics, it really shouldn't bring down the entire server. This is where `catch_panic` before calling each handler would be useful.

It would also be nice if the example-app handled cases where the value was not valid (>255, <0, or not an integer) with a 400 Bad Request extender to present a nice error page. In either case, i think a blank page (which happens now) is worse than causing the browser to show that it couldn't connect to the site or something.

smangelsdorf:
Attempting to catch panics from arbitrary futures requires a few steps:

1. Wrap the future in an unwind-safe wrapper which poisons the future if a panic occurs, to prevent further calls to `poll()` after such a panic.
2. Use the `catch_unwind` combinator from the `futures` crate to trap any panics from the now-unwind-safe `Future`.
3. Map the panic result into a hyper `Response`.

Fixes #17

### DIFF
--- a/src/handler/trap.rs
+++ b/src/handler/trap.rs
@@ -3,8 +3,11 @@
 
 use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::error::Error;
+use std::any::Any;
+use std::{io, mem};
 
 use hyper::{self, Response, StatusCode};
+use futures::Async;
 use futures::future::{self, Future, FutureResult};
 
 use handler::{NewHandler, Handler, HandlerError, IntoResponse};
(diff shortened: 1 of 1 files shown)


bradleybeddoes:
This looks fine, will hold merge until after I run it against ITP just to verify against large, real world Gotham client.

bradleybeddoes:
rebased with latest master, awaiting tooling success before merge.

PR #63, opened 2017-11-03
Update borrow-bag crate

ChristophWurst:
I've simplified the `BorrowBag` construction to idiomatic Rust code.

- [ ] Requires https://github.com/gotham-rs/borrow-bag/pull/8

Let me know what you think. This change seams not to break the public interface.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ mime = "0.3"
 futures = "~0.1.11"
 tokio-core = "0.1"
 mio = "0.6"
-borrow-bag = "0.3"
+borrow-bag = "0.4"
 url = "1.4.0"
 uuid = { version = "0.5", features = ["v4"] }
 chrono = "0.4"
--- a/src/router/route/dispatch.rs
+++ b/src/router/route/dispatch.rs
@@ -2,7 +2,7 @@
 
 use std::sync::Arc;
 use std::panic::RefUnwindSafe;
-use borrow_bag::{new_borrow_bag, BorrowBag, Handle, Lookup};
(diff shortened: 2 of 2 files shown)


ChristophWurst:
Of course, CI will fail now as the depending changes in the borrow-bag repo are not on crates.io.

ChristophWurst:
@bradleybeddoes @samsymons a general question: should I propose changes like this before sending a PR? I've seen a few spots in the code which could be easily refactored and I'd be happy to help developing this framework :)

smangelsdorf:
> should I propose changes like this before sending a PR?

Personally, I'm fine with small refactors like this. The biggest risk with a refactor is that the author has wasted their time if the PR is rejected (e.g. for taking the framework in an unwanted direction), so as a general rule any work where you'd be upset about the lost time should be discussed first.

Small, easy-to-review refactors are most welcome, especially if they make Gotham's code more idiomatic :+1:

bradleybeddoes:
Thanks for your work on Gotham so far @ChristophWurst, great to see other folks getting involved with the project.

Just wanted to add my two cents here (which essentially agrees with @smangelsdorf). I'm very happy to consider small pull requests, the kind of thing that would take an hour (at most two) to put together.

Anything that would take longer than that however a discussion issue or a discussion on the Gotham channel at https://gitter.im/gotham-rs/gotham would be much better suited. We're all contributing to open source for "free" so spending a small amount of effort on discussion/consensus before spending hours going down a path that may not work for the project for some (perhaps non obvious) reason is a much better use of everyones precious time.

PR #65, opened 2017-11-14
Refactor set_request_id into method of State

ChristophWurst:
This simplifies the exposed interface. IMO it makes sense to have this function as method.

### DIFF
--- a/src/handler/error.rs
+++ b/src/handler/error.rs
@@ -94,11 +94,10 @@ impl HandlerError {
     /// # use hyper::header::Headers;
     /// # use gotham::state::State;
     /// # use gotham::handler::{IntoResponse, IntoHandlerError};
-    /// # use gotham::state::request_id::set_request_id;
     /// # fn main() {
     /// # let mut state = State::new();
     /// # state.put(Headers::new());
-    /// # set_request_id(&mut state);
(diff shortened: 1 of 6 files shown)


ChristophWurst:
If we go down that path, `request_id` should be a method too. I put this on hold until I know whether this refactoring would be accepted :)

smangelsdorf:
We did debate this earlier in the project, and at the time arrived at the conclusion of keeping `State` a mostly "dumb" container. My main concern is that the `State` API would grow complex over time as we add more pieces of "blessed" state data, each of which gets a new method.

Bare functions seemed to me to be the less intrusive choice for general use, but then it's hard to say without a lot of general use to reference.

@bradleybeddoes thoughts on this change?

ChristophWurst:
Alright, makes sense!

ChristophWurst:
> We did debate this earlier in the project, and at the time arrived at the conclusion of keeping State a mostly "dumb" container. My main concern is that the State API would grow complex over time as we add more pieces of "blessed" state data, each of which gets a new method.

On a second thought, Rust's traits would be a perfect fit for this. You could keep `State` dead simple while extending its API using a trait like `RequestId` which would be implemented for the `State` struct.

I'm closing this PR now.

PR #66, opened 2017-11-15
Update hyper to 0.11.7

ChristophWurst:
This allows sharing of the tokio core handle by using [`hyper::server::Http::serve_addr_handle`](https://docs.rs/hyper/0.11.7/hyper/server/struct.Http.html#method.serve_addr_handle) to create the http
server.

Motivation: I'm currently trying to integrate [Sentry](https://github.com/SecurityInsanity/sentry-rs) as Gotham middleware to report errors/crashes. Since the library uses async Hyper, a handle reference is needed. However, with the previous version of Hyper, the handle was created together with the server. At that point the router had to be set up already and thus it was impossible to use the same tokio core. Fortunately, the Hyper lib seems to have added this feature of a shared core recently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ ci = []
 
 [dependencies]
 log = "0.3"
-hyper = "~0.11.2"
+hyper = "~0.11.7"
 serde = "~1.0"
 serde_derive = "~1.0"
 bincode = "0.8"
--- a/examples/kitchen-sink/Cargo.toml
+++ b/examples/kitchen-sink/Cargo.toml
@@ -8,7 +8,7 @@ authors = ["Shaun Mangelsdorf <s.mangelsdorf@gmail.com>",
 gotham = { path = "../.." }
 gotham_derive = { path = "../../gotham_derive" }
 futures = "*"
-hyper="~0.11.2"
+hyper="~0.11.7"
 borrow-bag = "*"
 log = "0.3"
 fern = "0.4"


ChristophWurst:
Inspired by https://github.com/hyperium/hyper/blob/39cf6ef7d26b3d829ec19fb1db176e8221170cb3/examples/multi_server.rs

my POC application code looks something like
```rust
pub fn main() {
    let addr = "127.0.0.1:7878".parse().unwrap();

    let mut core = Core::new().unwrap();
    let handle = core.handle();

    let credentials = SentryCredential::from_str("https://xxxxxxxx@sentry.io/123456").unwrap();
    let sentry = Sentry::new(
        core.handle().clone(),
        "Server Name".to_string(),
        "release".to_string(),
        "test_env".to_string(),
        credentials,
    );

    let server = Http::new()
        .serve_addr_handle(&addr, &handle, NewHandlerService::new(build_router(sentry)))
        .unwrap();

    println!("Listening on http://{}", server.incoming_ref().local_addr());

    let handle2 = handle.clone();
    handle.spawn(server.for_each(move |conn| {
        handle2.spawn(conn.map(|_| ()).map_err(|err| println!("server error: {:?}", err)));
        Ok(())
    }).map_err(|_| ()));

    core.run(futures::future::empty::<(), ()>()).unwrap();
}
```

ChristophWurst:
> The Travis CI build failed

Somehow Rust beta fails with `error: An unknown error occurred`.

ChristophWurst:
@smangelsdorf could you please restart the Travis build? I ran the tests locally with Rust beta and they succeeded.

bradleybeddoes:
I've kicked CI off again on beta channel

PR #67, opened 2017-11-18
Add an example using Tokio Timer

bradleybeddoes:
This example came up on #gotham and seemed useful as another basic example we could point to when using futures.


examples/kitchen-sink/src/main.rs
```diff
@@ -243,6 +252,31 @@ impl Echo {
         Box::new(future::lazy(move || future::ok((state, res))))
     }
 
+    pub fn wait(state: State) -> Box<HandlerFuture> {
+        let timeout = Timer::default();
+        let sleep = timeout.sleep(Duration::from_secs(2));
+
+        let result = sleep.then(|future_result| match future_result {
```
smangelsdorf: Nit: `FutureResult` is a type, but `future_result` is a `Result<_, _>`. Might be misleading for somebody who knows the futures library but is new to Gotham.

bradleybeddoes:
var renamed @smangelsdorf 

twas renamed sir

PR #70, opened 2017-12-06
Add experimental Router builder API

smangelsdorf:
This is a first attempt at building the `Router` creation API that will be used by all Gotham applications.

The primary goal of this branch is to get as far as we can _without_ introducing any macros. Keeping in line with idiomatic Rust, I opted to write a "builder" API for the router. An simple example router with this API looks like:

```rust
fn router() -> Router {
    // Note: Pipeline API isn't part of this PR
    let pipelines = new_pipeline_set();
    let (pipelines, default) =
        pipelines.add(new_pipeline().add(NewSessionMiddleware::default()).build());
    let pipelines = finalize_pipeline_set(pipelines);
    let default_pipeline_chain = (default, ());

    // Router builder starts here
    build_router(default_pipeline_chain, pipelines, |route| {
        route.get("/").to(welcome::index);

        route.get("/dashboard").to(dashboard::index);

        route.scope("/api", |route| {
            route.post("/submit").to(api::submit);
        });
    })
}
```

More in-depth examples:

* [The kitchen-sink example app has been updated to use this API](https://github.com/gotham-rs/gotham/blob/de5d8cd737bff37342f08d3e60886c1b705f4735/examples/kitchen-sink/src/main.rs#L78-L100)
* [Test case in `gotham::router::builder`](https://github.com/gotham-rs/gotham/blob/de5d8cd737bff37342f08d3e60886c1b705f4735/src/router/builder/mod.rs#L367-L393)
* Documentation available by running `cargo doc --features experimental-router` on this branch

This is an experimental API which will likely see some breaking changes before we're totally happy with it, ~~and so it will ship behind a Cargo feature (`"experimental-router"`) to communicate that fact to app developers who might otherwise expect some stability in the API.~~

As a future concern, once this API is bedded in and we know more about what we _can't_ do with a builder API, we can look at adding macros. That's out of scope for this PR.

Comments/questions/feedback/review welcome from anybody who's interested in doing so.

### DIFF
--- a/src/router/builder/mod.rs
+++ b/src/router/builder/mod.rs
@@ -1,4 +1,4 @@
-#![allow(warnings)]
+//! Defines a builder API for constructing a `Router`.
 
 mod draw;
 mod single;
@@ -6,22 +6,18 @@ mod replace;
 
 use std::marker::PhantomData;
 
-use hyper::Method;
-
-use handler::{Handler, NewHandler};
 use router::Router;
 use router::tree::TreeBuilder;
 use router::response::finalizer::ResponseFinalizerBuilder;
-use router::route::{Delegation, Extractors, RouteImpl};
-use router::route::matcher::{RouteMatcher, MethodOnlyRouteMatcher};
-use router::route::dispatch::{PipelineHandleChain, PipelineSet, DispatcherImpl};
-use router::request::path::{PathExtractor, NoopPathExtractor};
-use router::request::query_string::{QueryStringExtractor, NoopQueryStringExtractor};
-use router::tree::node::{SegmentType, NodeBuilder};
+use router::route::Delegation;
+use router::route::matcher::RouteMatcher;
+use router::route::dispatch::{PipelineHandleChain, PipelineSet};
+use router::request::path::PathExtractor;
+use router::request::query_string::QueryStringExtractor;
+use router::tree::node::NodeBuilder;
 
 pub use self::single::DefineSingleRoute;
 pub use self::draw::{DrawRoutes, DefaultSingleRouteBuilder};
-use self::replace::{ReplacePathExtractor, ReplaceQueryStringExtractor};
 
 /// Builds a `Router` using the provided closure. Routes are defined using the `RouterBuilder`
 /// value passed to the closure, and the `Router` is constructed before returning.
@@ -79,6 +75,8 @@ where
     Router::new(tree_builder.finalize(), response_finalizer)
 }
 
+/// The top-level builder which is created by `build_router` and passed to the provided closure.
+/// See the `build_router` function and the `DrawRoutes` trait for usage.
 pub struct RouterBuilder<'a, C, P>
 where
     C: PipelineHandleChain<P> + Copy + Send + Sync + 'static,
@@ -90,6 +88,8 @@ where
     response_finalizer_builder: ResponseFinalizerBuilder,
 }
 
+/// A scoped builder, which is created by `DrawRoutes::scope` and passed to the provided closure.
+/// See the `DrawRoutes` trait for usage.
 pub struct ScopeBuilder<'a, C, P>
 where
     C: PipelineHandleChain<P> + Copy + Send + Sync + 'static,
@@ -136,6 +136,8 @@ where
     }
 }
 
+/// Implements the traits required to define a single route, after determining which request paths
+/// will be dispatched here. The `DefineSingleRoute` trait has documentation for using this type.
 pub struct SingleRouteBuilder<'a, M, C, P, PE, QSE>
 where
     M: RouteMatcher + Send + Sync + 'static,
@@ -209,6 +211,7 @@ where
     }
 }
 
+// Trait impls live with the traits.
 impl<'a, M, C, P, PE, QSE> SingleRouteBuilder<'a, M, C, P, PE, QSE>
 where
     M: RouteMatcher
@@ -229,6 +232,8 @@ where
         + Sync
         + 'static,
 {
+    /// Coerces the type of the internal `PhantomData`, to replace an extractor by changing the
+    /// type parameter without changing anything else.
     fn coerce<NPE, NQSE>(self) -> SingleRouteBuilder<'a, M, C, P, NPE, NQSE>
     where
         NPE: PathExtractor + Send + Sync + 'static,
@@ -258,7 +263,7 @@ mod tests {
     use middleware::pipeline::new_pipeline;
     use middleware::session::NewSessionMiddleware;
     use state::{State, StateData};
-    use handler::{Handler, NewHandlerService};
+    use handler::NewHandlerService;
     use router::route::dispatch::{new_pipeline_set, finalize_pipeline_set};
     use router::response::extender::StaticResponseExtender;
     use router::tree::SegmentMapping;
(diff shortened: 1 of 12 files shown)


ChristophWurst:
IMHO this looks great. Well done!

ChristophWurst:
Btw, why exactly do you choose to hide this behind a feature flag? IMO, it's fine and expected to have breaking changes in a 0.x version. After all this framework is still pretty new and naturally its API will change a few times until it reaches a certain level of stability 😉

smangelsdorf:
> Btw, why exactly do you choose to hide this behind a feature flag?

For me it's more of a communicative tool, that we aren't just shipping this API and drawing a line under it. We **expect** it to change, and we're asking for feedback on it, especially after its initial release. 

That said, I see your point about 0.x already conveying that we want feedback and will make breaking changes. I suppose there would be no harm in just shipping it without the feature flag, but I'm very much on the fence about it. @bradleybeddoes thoughts?

bradleybeddoes:
I was happy to go with the feature flag because this is something that is very much in flux, whereas the rest of Gotham we hope is a little more stable/future proofed (naturally we are using 0.x so we can change if a better idea comes along).

That being said @ChristophWurst does have a valid view point.

As we're on the fence I suggest the Gotham community should help us out here, please vote!

[](https://api.gh-polls.com/poll/01C0Z06JQZKT1T5HP32V6ZKNEH/Feature%20Flag/vote)
[](https://api.gh-polls.com/poll/01C0Z06JQZKT1T5HP32V6ZKNEH/No%20Feature%20Flag/vote)
[](https://api.gh-polls.com/poll/01C0Z06JQZKT1T5HP32V6ZKNEH/Candy/vote)

rcaught:
IMO people can always pin to a version if it breaks for them (which should be very apparent since it's Rust).

Maybe the original intent of semver 0.x is that you don't want to stifle rapid innovation and overcomplicate the source, so I vote for no feature flag.

smangelsdorf:
At the time of writing, the poll results are unanimous on removing the feature flag, so that's exactly what I've done with 626ac8f.

src/router/builder/replace.rs
```diff
@@ -0,0 +1,60 @@
+use std::panic::RefUnwindSafe;
+
+use router::request::path::PathExtractor;
+use router::request::query_string::QueryStringExtractor;
+use router::route::matcher::RouteMatcher;
+use router::route::dispatch::PipelineHandleChain;
+use router::builder::SingleRouteBuilder;
+use router::builder::single::DefineSingleRoute;
+
+pub trait ReplacePathExtractor<T>
```
bradleybeddoes: Will need to add documentation here

src/router/builder/replace.rs
```diff
@@ -0,0 +1,60 @@
...
+    NPE: PathExtractor + Send + Sync + 'static,
+{
+    type Output = SingleRouteBuilder<'a, M, C, P, NPE, QSE>;
+
+    fn replace_path_extractor(self) -> Self::Output {
+        self.coerce()
+    }
+}
+
+pub trait ReplaceQueryStringExtractor<T>
```
bradleybeddoes: Will need to add documentation here

src/router/builder/single.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// #
+    /// #   let default_pipeline_chain = (default, ());
+    ///
+    /// build_router(default_pipeline_chain, pipelines, |route| {
+    ///     route.get("/request/path").to(my_handler);
+    /// })
+    /// # }
+    /// # fn main() { router(); }
+    /// ```
+    fn to<H>(self, handler: H)
```
bradleybeddoes: This is purely discussion.

I am wondering if the naming here should be `to/to_handler` as opposed to `to/to_new_handler` as I could easily see folks skipping over the concept of the `NewHandler` type (at least initially) and this getting confusing.

Brevity in the builder is also a nice side effect of this if we decided to change.

I understand this makes the builder not quite "pure" in naming, but worth considering.

src/router/builder/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+///
+///     let default_pipeline_chain = (default, ());
+///
+///     build_router(default_pipeline_chain, pipelines, |route| {
+///         route.get("/request/path").to(my_handler);
+///     })
+/// }
+/// # fn main() { router(); }
+/// ```
+pub fn build_router<C, P, F>(pipeline_chain: C, pipelines: PipelineSet<P>, f: F) -> Router
```
bradleybeddoes: I appreciate this PR has not considered pipelines so I'm happy for this comment to be considered in the future if overkill now.

I was considering use cases for Gotham and it strikes me that `pipelines` could very easily be something that developers who just want something simple could do without. Likewise in most of our API example documentation, where it becomes ~9 lines of boilerplate that could be left out.

How would you feel about adding a function like: 

`pub fn build_simple_router<F>(f: F) -> Router`

That internally does the basic pipeline setup before delegating to `build_router`? 

Moving between the two, if developers have some future use case for pipelines come up, would be a simple exercise as well.

bradleybeddoes:
So i've done a solid first read through of this PR and left some comments. 

In the next day or so I'll start putting some real world code against this PR, see if I can find any funky edge cases we may need to consider.

src/router/builder/single.rs
smangelsdorf: I disagree, for the same reason you mentioned. Though, instead of calling it "pure", I'd just call it "accurate".

`Handler + Copy` is the "default" here, which completely obviates the need for `NewHandler` in the trait bounds for `to<H>`. Having `to` take a `Handler + Copy`, and `to_handler` take a `NewHandler` seems like it would be _far_ more confusing for people who've learned what a `NewHandler` is.

For the 99.9% case, and especially new Gotham users, I see `to_new_handler` not being used at all. It's just there for completeness, in being able to control the exact `NewHandler` which is used for the route.

src/router/builder/mod.rs
smangelsdorf: Sounds fine to me. I assume by "basic pipeline setup" you mean with no middleware?

bradleybeddoes: Indeed, as bare bones as it can be, essentially "just give me routes to my functions".

src/router/builder/single.rs
bradleybeddoes: I can live with this reasoning, consider this resolved.

bradleybeddoes:
@smangelsdorf I've pushed a branch with changes to support sub-routers which can be compared at https://github.com/gotham-rs/gotham/compare/router-builder...test-helpers-and-new-tree?expand=1 (includes test helper changes due to what I am doing locally) or looked at as a standalone patch at https://gist.github.com/bradleybeddoes/698dc5212a32a919d1950800ac6402fa (does not include test helper changes).

If you can consider this addition and patch your branch if you're ok with the approach here that would be very helpful.

I was also looking at the support for `get` and `post` within `DrawRoutes`. Ideally we'd have these for all the common HTTP verbs (this may be the intent and this is currently just PoC but I thought it was worth mentioning).

Now that I have support for sub routers I'll continue migrating by extensive router code into the builder API and provide further feedback here.

bradleybeddoes:
ping @smangelsdorf

smangelsdorf:
@bradleybeddoes I made a slight tweak to the approach you took, though I think you'll find it's fundamentally the same idea. Rather than having the single route builder deal with delegated routes, they're done via a different piece of API like:

```rust
build_router(default_pipeline_chain, pipelines, |route| {
    route.delegate("/delegated").to_router(delegated_router);
});
```

The reasoning is that app devs are prevented at compile time from doing odd things like specifying path/query extractors and route matchers* when they're delegating to another router.

Thoughts, etc welcome. If I've missed something that your version could do, happy to tweak.

\* There's a very small possibility that someone will want to use a route matcher to guard a subtree, but I don't think we should be tackling that piece of API now. I'd find it very odd if they were using method matchers in particular, and that's pretty much all we have.

smangelsdorf:
I'll add support for the other HTTP verbs. We've come this far, so it probably makes sense to just blow the line count out to support all the verbs 😁

src/router/builder/mod.rs
smangelsdorf: For posterity: Sorted in 685e0d5

smangelsdorf:
All feedback has been addressed now. Awaiting further review.

PR #75, opened 2017-12-20
Proposal to remove lifetime from TestResponse

bradleybeddoes:
In 3rd party code it may be useful to setup a TestServer and TestResponse within a common function, returning both to the the test harness for further evaluation.

This is certainly a pattern I use a lot within my tests, that was "broken" by the recently merged changes to Gotham testing.

Example that is possible if this PR is merged that currently isn't on master:

``` rust
fn send_request(uri: String, accept: Option<mime::Mime>) -> (TestServer<Router>, TestResponse) {
    env::set_var("APP_ENV", "test");
    let mut test_server = TestServer::new(router()).unwrap();
    let mut req = Request::new(Method::Get, uri.parse().unwrap());
    match accept {
        Some(accept) => req.headers_mut().set(Accept(vec![qitem(accept)])),
        None => (),
    };
    let res = test_server
        .client()
        .perform(req)
        .unwrap();

    (test_server, res)
}

#[test]
fn valid_abc() {
    let subject = format!("acct:{}", fake!(Internet.free_email));
    let (_test_server, res) = send_request(web_finger_uri(&subject, OIDC_DISCOVERY_REL), None);
    
    // asserts
}

#[test]
fn valid_xyz() {
    let subject = format!("acct:{}", fake!(Internet.free_email));
    let (_test_server, res) = send_request(web_finger_uri(&subject, OIDC_DISCOVERY_REL), Some(custom_mime_type());
    
   // asserts
}
```

This hasn't been discussed, prior to my landing of this PR, as being a good idea or not so lets do that now :grin:.

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -58,10 +58,7 @@ where
     data: Rc<TestServerData<NH>>,
 }
 
-/// This type should not need to be directly used by client applications.
-///
-/// It is used by TestServer to hold internal data within an Rc
-pub struct TestServerData<NH = Router>
+struct TestServerData<NH = Router>
 where
     NH: NewHandler + 'static,
 {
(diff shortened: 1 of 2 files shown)


smangelsdorf:
Sounds like a good idea. I'll save the proper review for a time of day where I'm more alert 😁

Issue #74, opened 2017-12-19
Simple Middleware support


nyarly:
In my freshman app with Gotham I have:

```rust
mod app_config {
  use gotham::middleware::{NewMiddleware, Middleware};
  use gotham::state::State;
  use hyper::Request;
  use gotham::handler::HandlerFuture;
  use std::io;
  use std::env;
  use state::AppConfig;
  use futures::{future, Future};

  pub struct New {}

  impl NewMiddleware for New {
    type Instance = Ware;
    fn new_middleware(&self) -> io::Result<Self::Instance> {
      Ok(Ware {})
    }
  }

  pub struct Ware {}

  impl Middleware for Ware {
    fn call<Chain>(self, mut state: State, request: Request, chain: Chain) -> Box<HandlerFuture>
      where Chain: FnOnce(State, Request) -> Box<HandlerFuture> + Send + 'static,
            Self: Sized
    {
      let cfg = AppConfig {
        canonical_url: env::var("CANONICAL_URL").unwrap_or_default(),
        // ...
      };

      debug!("AppConfig: putting config in state");
      state.put(cfg);
      Box::new(future::ok(state).and_then(|state| chain(state, request)))
    }
  }
}
```

The only thing of interest there is the contents of the call method. It would be nice to only need to write/register that.

Could Middleware/NewMiddleware be implmented for a Fn trait?

This may be a simple as a macro, but I don't have the scope within Gotham to speak to the required generality.

smangelsdorf:
Thanks for raising this. The `NewMiddleware` / `Middleware` split is not ideal in the very simple case, but is necessary to support anything more complex and especially something which involves a connection pool.

One way to handle this is by deriving `Clone` and then implementing `NewMiddleware` + `Middleware` like:

```rust
#[derive(Clone)]
struct MyType;

impl NewMiddleware for MyType {
  type Instance = Self;
  fn new_middleware(&self) -> io::Result<Self> {
    self.clone()
  }
}

impl Middleware for MyType {
  // ... implementation elided
}
```

We could make this easier by adding a custom derive for `NewMiddleware`, so the simple middleware definition could just be:

```rust
#[derive(Clone, NewMiddleware)]
struct MyType;

impl Middleware for MyType {
  // ... implementation elided
}
```

> Could Middleware/NewMiddleware be implmented for a Fn trait?

[...]

PR #83, opened 2018-01-09
Implement custom derive for NewMiddleware

smangelsdorf:
As raised in #74, it's often the case that the `NewMiddleware` implementation is just a piece of boilerplate. A lot of use cases for middleware can be implementing using a single struct which just uses `Clone` to implement `NewMiddleware`. This goes one step further and allows `NewMiddleware` to be derived too.

The derived `NewMiddleware` implementation requires that the type also implements `Clone`.

### DIFF
--- a/examples/kitchen-sink/src/middleware.rs
+++ b/examples/kitchen-sink/src/middleware.rs
@@ -1,9 +1,7 @@
-use std::io;
-
 use gotham;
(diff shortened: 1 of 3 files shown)


bradleybeddoes:
CI issues are noted and due to changes in a dependency. Fault is not relevant to this PR so continuing with merge, PR to address root cause being created.

Issue #78, opened 2017-12-30
API design to get hold of a Tokio handle within a handler


ChristophWurst:
As discussed on Gitter with @bradleybeddoes, I'm missing the possibility to send new HTTP requests inside Gotham handlers because currently it's not possible to get a reference to the underlying Tokio event loop. 

@bradleybeddoes pointed me to https://github.com/gotham-rs/gotham/blob/953c39b6d5d5134bbab5ab658f1aadc243a68806/src/handler/mod.rs#L151 where other framework state variables are initialized. Unfortunately, I don't see a way of accessing the tokio handle in that function. As a matter of fact the whole Gotham codebase is pretty much unaware of the whole Tokio stuff because the way Gotham apps seem to be started is that the handler is passed to Hyper's server API. Hyper will either create the Tokio core internally, or let you use an existing one (since https://github.com/gotham-rs/gotham/pull/66).

An alternative to having Gotham put the handle reference into the state object would be to use a middleware for that: https://github.com/ChristophWurst/gotham-middleware-tokio. You can find an example usage here: https://github.com/ChristophWurst/gotham-middleware-tokio/blob/master/examples/hyper_client.rs#L22

There seem to be a few problems though. I don't know whether we should use a `Handle`, a `Remote` or let the developer choose. `Handle` cannot be shared across threads. However, Hyper expects a `Handle` rather than a `Remote` for its client, making it a bit cumbersome to use. The remote object does have method to get a handle, [although it's not guaranteed to always get one](https://docs.rs/tokio-core/0.1.11/tokio_core/reactor/struct.Remote.html#method.handle).

This is somewhat related to https://github.com/hyperium/hyper/issues/1002

@bradleybeddoes @smangelsdorf thoughts on how to integrate this in Gotham?

bradleybeddoes:
I'm not sure what @smangelsdorf had in mind but I was thinking of something along the lines of passing a handle reference to https://github.com/gotham-rs/gotham/blob/953c39b6d5d5134bbab5ab658f1aadc243a68806/src/handler/mod.rs#L123 and then  being able to use that within https://github.com/gotham-rs/gotham/blob/953c39b6d5d5134bbab5ab658f1aadc243a68806/src/handler/mod.rs#L151 to push it into `state`.

I may be missing some importance nuance here as I've only very briefly considered this, happy to keep kicking this around if I've overlooked something important.

[...]

PR #89, opened 2018-01-11
Address Hyper deprecations and provide access to Tokio Handle

smangelsdorf:
This refactors some Gotham internals to address deprecation warnings in the latest version of Hyper. As a part of this refactor I've addressed #78 by passing the Tokio `Handle` through into `State` too.

This was mostly written before #84 and then merged with `master`. I was pretty careful in addressing the merge conflicts, but I think it's worth mentioning so you can keep an eye out for anything that's out of place.


gotham/src/clone_sockets.rs
```diff
@@ -30,30 +31,35 @@ where
...
-    new_service: &NewHandlerService<NH>,
-) where
+fn serve<NH>(listener: TcpListener, addr: &SocketAddr, protocol: &Http, new_handler: Arc<NH>)
+where
     NH: NewHandler + 'static,
 {
     let mut core = Core::new().expect("unable to spawn tokio reactor");
     let handle = core.handle();
 
+    let new_service = GothamService::new(new_handler, handle.clone());
```
bradleybeddoes: Nit: I feel it may help readability to `s/new_service/gotham_service`.

Thoughts?

gotham/src/service/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+    fn clone(&self) -> Self {
+        ConnectedGothamService {
+            t: self.t.clone(),
+            handle: self.handle.clone(),
+            client_addr: self.client_addr,
+        }
+    }
+}
+
+impl<T> NewService for ConnectedGothamService<T>
```
bradleybeddoes: I've looked into the types and I believe I've convinced myself that `ConnectedGothamService` does not actually need to be `NewService` any longer, which is turn simplifies code in `clone_sockets.rs` and various test cases.

Is there any benefit to this remaining in place I may not be considering?

gotham/src/clone_sockets.rs
bradleybeddoes: Likewise https://github.com/gotham-rs/gotham/blob/481a912490881f3c9016db6a48d2a53c00406c41/gotham/src/socket_queue.rs#L109

gotham/src/service/mod.rs
bradleybeddoes: Example of the code I hacked on with `NewService` removed if useful:

``` rust
    core.run(listener.incoming().for_each(|(socket, addr)| {
        let service = new_service.connect(addr);
        let f = protocol.serve_connection(socket, service).then(|_| Ok(()));
        //.map(|_| ())
        //.map_err(|_| ());

        handle.spawn(f);
        Ok(())
    })).expect("unable to run reactor over listener");
```

smangelsdorf:
Good ideas. That tidied things up nicely.

PR #97, opened 2018-01-15
Add PUT, PATCH and DELETE to TestServer

bradleybeddoes:
I'd like **put** this **patch** into Gotham that adds functionality and doesn't **delete** anything.

Oh yes, that just happened.

### DIFF
--- a/gotham/src/test/mod.rs
+++ b/gotham/src/test/mod.rs
@@ -244,6 +244,11 @@ impl<NH> TestClient<NH>
 where
     NH: NewHandler + 'static,
 {
+    /// Parse the URI and begin constructing a HEAD request using this `TestClient`.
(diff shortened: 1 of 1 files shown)


gotham/src/test/mod.rs
```diff
@@ -244,6 +244,11 @@ impl<NH> TestClient<NH>
 where
     NH: NewHandler + 'static,
 {
+    /// Parse the URI and begin constructing a HEAD request using this `TestClient`.
+    pub fn head(self, uri: &str) -> RequestBuilder<NH> {
+        self.build_request(Method::Head, uri)
+    }
```
smangelsdorf: `head_uri` omitted deliberately?

PR #100, opened 2018-01-16
Include `Allow` header with a 405 response

smangelsdorf:
This populates each `Node` in the router tree with a list of HTTP methods that should be sent in an `Allow` header, if a request reaches that `Node` but is rejected with `Err(StatusCode::MethodNotAllowed)`.

It's a different approach to what's described in #3, and if there is some nuance that is missed by this implementation I'm more than happy to discuss / refine / rewrite to make sure we have the correct behaviour. However, all the cases I can think of are covered (particularly, if a correctly-implemented `RouteMatcher` is rejecting a request with a `405` response, the `Allow` list should be correct).

The goal in taking this approach was to avoid mutating `State` during the normal course of serving a request, to populate it with data that won't be used for the vast majority of requests.

### DIFF
--- a/gotham/src/router/mod.rs
+++ b/gotham/src/router/mod.rs
@@ -11,6 +11,7 @@ use std::sync::Arc;
 
 use futures::{future, Future};
 use hyper::{Response, StatusCode};
+use hyper::header::Allow;
 
 use handler::{Handler, HandlerFuture, IntoResponse, NewHandler};
 use http::request::path::RequestPathSegments;
@@ -108,7 +109,13 @@ impl Handler for Router {
                         },
                         Err(status) => {
                             trace!("[{}] responding with error status", request_id(&state));
-                            let res = create_response(&state, status, None);
+                            let mut res = create_response(&state, status, None);
+
+                            if let StatusCode::MethodNotAllowed = status {
+                                res.headers_mut()
+                                    .set(Allow(leaf.allow_header_method_list()));
+                            }
+
                             Box::new(future::ok((state, res)))
                         }
                     }
@@ -214,6 +221,7 @@ mod tests {
     use router::route::dispatch::{finalize_pipeline_set, new_pipeline_set, DispatcherImpl};
     use router::route::matcher::MethodOnlyRouteMatcher;
     use router::response::finalizer::ResponseFinalizerBuilder;
+    use router::builder::*;
     use state::set_request_id;
     use handler::HandlerError;
 
@@ -259,6 +267,25 @@ mod tests {
             }
             Err(_) => panic!("Router should have handled request"),
         };
+    }
+
+    #[test]
+    fn allow_header_if_method_not_allowed() {
+        let router = build_simple_router(|route| {
+            route.get_or_head("/test").to(handler);
+            route.delete("/test").to(handler);
+            route.options("/test/2").to(handler);
+        });
+
+        let (_state, res) = send_request(router, Method::Options, "https://test.gotham.rs/test")
+            .map_err(|_| ())
+            .unwrap();
+
+        assert_eq!(StatusCode::MethodNotAllowed, res.status());
+        assert_eq!(
+            vec![Method::Delete, Method::Get, Method::Head],
+            **res.headers().get::<Allow>().unwrap()
+        )
     }
 
     #[test]
--- a/gotham/src/router/mod.rs
+++ b/gotham/src/router/mod.rs
@@ -273,6 +273,7 @@ mod tests {
     fn allow_header_if_method_not_allowed() {
         let router = build_simple_router(|route| {
             route.get_or_head("/test").to(handler);
+            route.get("/test").to(handler); // Proves deduplication works.
             route.delete("/test").to(handler);
             route.options("/test/2").to(handler);
         });
--- a/gotham/src/router/tree/node.rs
+++ b/gotham/src/router/tree/node.rs
@@ -3,7 +3,7 @@
 use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::borrow::Borrow;
-use hyper::StatusCode;
+use hyper::{Method, StatusCode};
 
 use http::PercentDecoded;
 use router::route::{Delegation, Route};
@@ -103,6 +103,10 @@ pub struct Node {
 
     routes: Vec<Box<Route + Send + Sync>>,
 
+    // The list of methods to send in an `Allow` header, when a request that reaches this node
+    // results in a `405 Method Not Acceptable` response.
+    allow_header_method_list: Vec<Method>,
+
     delegating: bool,
     children: Vec<Node>,
 }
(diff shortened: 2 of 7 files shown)


gotham/src/router/mod.rs
```diff
@@ -261,6 +269,25 @@ mod tests {
         };
     }
 
+    #[test]
+    fn allow_header_if_method_not_allowed() {
+        let router = build_simple_router(|route| {
+            route.get_or_head("/test").to(handler);
```
bradleybeddoes: Worth adding a `route.get("/test").to(handler)` here as well? Will prove the dedup is occurring during finalisation.

gotham/src/router/tree/node.rs
```diff
@@ -385,11 +393,22 @@ impl NodeBuilder {
...
 
+        let mut allow_header_method_list: Vec<Method> = self.routes
+            .iter()
+            .flat_map(|r| r.allow_header_method_list())
+            .collect();
+
+        allow_header_method_list
+            .sort_by(|a, b| AsRef::<str>::as_ref(a).cmp(AsRef::<str>::as_ref(b)));
+
+        allow_header_method_list.dedup();
```
bradleybeddoes: I was re-reading https://tools.ietf.org/html/7231#section-7.4.1

> The actual set of allowed methods is defined by the origin server at
   the time of each request.  An origin server MUST generate an Allow
   field in a 405 (Method Not Allowed) response and MAY do so in any
   other response.  An empty Allow field value indicates that the
   resource allows no methods, which might occur in a 405 response if
   the resource has been temporarily disabled by configuration.

Which raised 2 thoughts:

1. In the case where Gotham is matching requests but has not had any methods explicitly set, should we be returning all HTTP verbs instead of the empty set
2. Allow is a potential header for non 405 responses as well, probably worth altering associated comments to remove implication that this is only applicable for 405.

gotham/src/router/mod.rs
```diff
@@ -108,7 +109,13 @@ impl Handler for Router {
                         },
                         Err(status) => {
                             trace!("[{}] responding with error status", request_id(&state));
-                            let res = create_response(&state, status, None);
+                            let mut res = create_response(&state, status, None);
+
+                            if let StatusCode::MethodNotAllowed = status {
```
bradleybeddoes: I'm half wondering if this should be generic in some capacity `Node` would have a mapping on headers to set for each status code, might be overkill...

bradleybeddoes:
A few minor comments/thoughts but overall this looks good, not altering State is 👍

gotham/src/router/tree/node.rs
smangelsdorf: > In the case where Gotham is matching requests but has not had any methods explicitly set, should we be returning all HTTP verbs instead of the empty set

If that's the case, the `RouteMatcher` shouldn't be returning a `405 Method Not Allowed`, and the `Allow` header code won't be triggered. If a `RouteMatcher` permits all methods and then returns a 405, that's a whole 'nother problem 😁 

> Allow is a potential header for non 405 responses as well, probably worth altering associated comments to remove implication that this is only applicable for 405.

Perhaps, but that's the only way it's used *currently*, and it's populated with only that usage in mind. Other use cases would probably be incorrect due to the first point, and the fact that it's no longer only triggered by a 405 response.

gotham/src/router/mod.rs
smangelsdorf: The thought definitely crossed my mind, but seemed like over-engineering and something we could add later if there's another case.

If you feel strongly about it, I can address it now. No big deal.

gotham/src/router/mod.rs
smangelsdorf: Added.

gotham/src/router/tree/node.rs
smangelsdorf: > > In the case where Gotham is matching requests but has not had any methods explicitly set, should we be returning all HTTP verbs instead of the empty set
>
> If that's the case, the `RouteMatcher` shouldn't be returning a `405 Method Not Allowed`, and the `Allow` header code won't be triggered. If a `RouteMatcher` permits all methods and then returns a 405, that's a whole 'nother problem 😁 

To look at this another way:

```
GET / HTTP/1.1
Host: x.example.com
```

```
HTTP/1.1 405 Method Not Allowed
Allow: GET, HEAD, POST, PATCH, ... [all methods]
```

Doesn't make sense.

bradleybeddoes: I may have been less than clear in my description. The intent was to take this back to being generic, as it seems to me that it actually is, aside from populating all methods if `allow_header_method_list` is empty.

Right now the router check is as follows:

```
if let StatusCode::MethodNotAllowed = status ...
```

Which should only be the response status (custom implementations doing stupid things aside) if `MethodOnlyRouteMatcher` decides the request is not satisfactory. To do that it follows that `allow_header_method_list` must have been populated during tree construction, hence the 405 header would be correctly populated and your example above would not occur.

For folks that wanted to populate an `Allow` header for some other purpose `Allow(leaf.allow_header_method_list())` would then be just as applicable. Albeit accessing that API from something other than the router would be hard(impossible? different discussion...) for now.

I'm not bound to this, I just think there is an opportunity here to make this code more widely applicable (but likely only ever used by 405 in reality).

gotham/src/router/mod.rs
bradleybeddoes: Lets wait for a use case, consider this resolved for now.

gotham/src/router/tree/node.rs
smangelsdorf: > I'm not bound to this, I just think there is an opportunity here to make this code more widely applicable (but likely only ever used by 405 in reality).

I think that altering it to be more widely applicable is a minefield of possibilities. For example:

```
/a [MethodOnlyRouteMatcher - GET, POST]
/a [AnyRouteMatcher]
```

The `Allow` list in that case would only be populated with `[GET, POST]`. The fact that this list is strictly incorrect is irrelevant to the implementation because the HTTP method is always matched by the second route, so the `405` never fires.

The logic makes lists that are applicable for creating `Allow` headers for `405` response, and not always applicable for other purposes. Rather than try to engineer it to be correct in all cases, I just documented it as only being for `405` and left it as-is.

Making it applicable for all cases doesn't seem doable without generating the list on a per-request basis. That's the implementation I started with (the `Result<(), StatusCode>` became a `Result<(), NonMatch>` with appropriate metadata being propagated in the `NonMatch` enum), but dropped back to this when the change proved to be quite far-reaching.

If we want the ability to extend the usage of the `Allow` header later, I think we'll need to go back to this version of it. The question is whether we need it.

It will make the implementation more complex, but also more internally consistent.

Maybe it's a good idea anyway.

smangelsdorf:
Closing in favour of #107

PR #102, opened 2018-01-20
Remove log crate calls from derived extractors

### LINKED ISSUE #79: gotham_derive has an invisible dependency on the "log" crate
When trying to use gotham in a new project that happens to be using the "slog" crate for logging, I started getting incomprehensible "format argument must be a string literal." errors when trying to derive PathExtractor on a struct.

```rust
#[derive(StateData, PathExtractor, StaticResponseExtender)]
struct GetThreadRequestPath {
    no: i64,
}
```

After some experimentation, I discovered that the trigger for this error is having `#[macro_use] extern crate slog;` in the main file.

After having looked into it some more, I eventually found the source of the problem. The derive macros in "extractors.rs" in gotham_derive contain several instances of the `trace!` and `error!` macros from the "log" crate even though it doesn't even have the "log" crate as a dependency itself. Commenting those out in gotham_derive gets rid of the errors.

It seems like macros inside macros don't get expanded and just leak out into the surrounding code, causing it to basically have an "invisible" dependency on the log crate and causing it to fail with complaints about missing macros if the user does not have the "log" crate as a dependency and `#[macro_use] extern crate log;` in main. Or, as in my case, if you have another logging crate with incompatible versions of those macros, just give a completely uninformative compile error.

I'm not 100% sure what the best solution here is, but it seems to me like having macros inside macros is a bad idea, as it basically leaks invisible dependencies into the project using them.

Link to reproduction:
https://github.com/forbjok/gotham-repro

bradleybeddoes:
This was only providing trace output in the general case. 

For the error case the error returned has been slightly enhanced to sure information is not lost (The benefits of the parse return type should be considered at some point using String is not ideal).

Closes #79.

### DIFF
--- a/gotham/src/router/builder/single.rs
+++ b/gotham/src/router/builder/single.rs
@@ -150,8 +150,6 @@ pub trait DefineSingleRoute {
     /// # #[macro_use]
     /// # extern crate gotham_derive;
     /// # extern crate hyper;
-    /// # #[macro_use]
-    /// # extern crate log;
     /// # use hyper::Response;
     /// # use gotham::state::{State, FromState};
     /// # use gotham::router::Router;
@@ -206,8 +204,6 @@ pub trait DefineSingleRoute {
     /// # #[macro_use]
     /// # extern crate gotham_derive;
     /// # extern crate hyper;
-    /// # #[macro_use]
-    /// # extern crate log;
     /// # use hyper::Response;
     /// # use gotham::state::{State, FromState};
     /// # use gotham::router::Router;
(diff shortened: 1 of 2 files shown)


bradleybeddoes:
cc @forbjok and @onelson for feedback/real world testing .

onelson:
With a slight tweak to the hyper version in @forbjok's repro, it builds successfully. Looks good to me. :confetti_ball:

forbjok:
I updated the repro repository, and my other project using Gotham.
Seems to be working fine.

Issue #96, opened 2018-01-15
Router builder tweaks


bradleybeddoes:
Following some usage of the Router builder in "the real world" some tweaks have become clear:

Highlight in documentation:

* `scope` - Subtree in the same router, applies active `PipelineChain`
* `delegate` - Delegates a whole subtree to a secondary `Router` instance, applies active `PipelineChain`. Delegated `Router` may define and apply own `PipelineChain` instances as normal.

Add:

* `delegate_without_pipelines` - Delegates a whole subtree to a secondary `Router` instance, **does not apply** active `PipelineChain`. Delegated `Router` may define and apply own `PipelineChain` instances as normal.
* `associate` - Pins to a point in the tree and defines multiple routes, e.g.

```
route.post("/address").to(checkout::address::create);
route.put("/address").to(checkout::address::update);
route.patch("/address").to(checkout::address::update);
route.delete("/address").to(checkout::address::delete);
```

becomes something like:

```
route.associate("/address", |assoc| {
    assoc.post().to(checkout::address::create);
    assoc.put().to(checkout::address::update);
    assoc.patch().to(checkout::address::update); 
    assoc.delete().to(checkout::address::delete);
})
```

PR #105, opened 2018-01-21
Implement associated routes and delegate_without_pipelines

smangelsdorf:
As described in #96.


gotham/src/router/builder/mod.rs
```diff
@@ -267,6 +267,382 @@ where
...
+    /// build_simple_router(|route| {
+    ///     route.associate("/resource/:id", |assoc| {
+    ///         let mut assoc = assoc.with_path_extractor::<MyPathExtractor>();
+    ///         assoc.get().to(handler);
+    ///     });
+    /// })
+    /// # }
+    /// # fn main() { router(); }
+    /// ```
+    pub fn with_path_extractor<'b, NPE>(&'b mut self) -> AssociatedRouteBuilder<'b, C, P, NPE>
```
bradleybeddoes: Should this also have a `with_query_extractor`? If so I am wondering if a Trait that could be applied here and to `DefineSingleRoute` may be useful if we want to support extra extractors in the future (e.g. body extractors, say x-www-form-urlencoded).

gotham/src/router/builder/draw.rs
```diff
@@ -450,3 +571,94 @@ where
...
+    use handler::HandlerFuture;
+    use middleware::{Middleware, NewMiddleware};
+    use state::State;
+    use router::builder::*;
+    use pipeline::*;
+    use pipeline::single::*;
+    use http::response::create_response;
+    use test::TestServer;
+
+    #[derive(Clone, Copy)]
```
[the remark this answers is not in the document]
smangelsdorf: Can't derive here because of the circular dependency it would create to depend on `gotham_derive`, unless I'm mistaken.

gotham/src/router/builder/mod.rs
smangelsdorf: Arguably yes, arguably no. I'm tending towards "no" because associated routes are bound to a _path_, and so the _path_ extractor makes sense to specify at the top level. Though as I say that, I imagine that a ticket will show up in a few months suggesting that its omission was a silly oversight.

My rule of thumb with factoring things into traits in the router builder is that it can't make the API documentation **harder** to read, so I'm not opposed as long as it fits the bill there. If a little duplication makes the type soup more palatable, I'm tending to err on that side.

gotham/src/router/builder/mod.rs
```diff
@@ -267,6 +267,382 @@ where
...
+    /// # fn router() -> Router {
+    /// build_simple_router(|route| {
+    ///     route.associate("/resource", |assoc| {
+    ///         assoc.request(vec![Method::Get, Method::Head, Method::Post]).to(handler);
+    ///     });
+    /// })
+    /// # }
+    /// # fn main() { router(); }
+    /// ```
+    pub fn request<'b>(&'b mut self, methods: Vec<Method>) -> DefaultSingleRouteBuilder<'b, C, P> {
```
smangelsdorf: Self-review: Might be useful if the implementation actually **used** the `PE` type parameter.

😳

bradleybeddoes:
General comment, this has CI errors when rebased on latest master.

```
---- src/router/builder/mod.rs - router::builder::AssociatedRouteBuilder<'a, C, P, PE>::with_path_extractor (line 294) stdout ----
	error: unused `#[macro_use]` import
 --> src/router/builder/mod.rs:2:1
  |
2 | #[macro_use]
  | ^^^^^^^^^^^^
  |
```

which is:

```
    /// ```rust
    /// # #[macro_use]
    /// # extern crate log;
...
```

Happy to fix in my branch if you don't want to rebase/address here.

gotham/src/router/builder/draw.rs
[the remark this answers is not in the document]
bradleybeddoes: :tired_face: carry on, nothing useful to be gleaned from this dumb suggestion.

gotham/src/router/builder/mod.rs
bradleybeddoes: No worries re: type soup/Traits and readability, reasonable trade offs.

wrt `with_query_extractor` I see your point the `Struct` being extracted into, if we leave it at the `route` level, could certainly just employ `Option` fields to make it suitable for each associated route, which is ok but not without draw backs.

I did consider however that this is something that could be achieved if you simply wrote out multiple `route.verb("/path").with_query_extractor...`. For that reason alone I felt it made sense to have it here when I was looking this over.

I could come across to the "no" argument but flexibility in the router builder, where it is reasonable API/code wise, seems like a good "goal".

gotham/src/router/builder/mod.rs
bradleybeddoes: :beer:

gotham/src/router/builder/mod.rs
smangelsdorf: Doesn't hurt to have it but also have it rarely used. I'll add it.

bradleybeddoes:
cherry pick 37c4d64 if you like to resolve fails on PR builds

PR #107, opened 2018-01-23
Include `Allow` header with a 405 response (rewrite)

smangelsdorf:
This is a competing solution for #3 (the other solution being #100). This builds the `Allow` list on-demand, rather than prepopulating it in the tree.

The advantages of this approach:

* We're only creating data which represents reality (contrast with #100 which unnecessarily populated a list)
* This is _far_ more capable of handling weird behaviour that could be thrown at us by custom `RouteMatcher` implementations in the future. I can think of some crazy `RouteMatcher` implementations that would trip up the logic in #100.
* The `Allow` list in this case could be used for _other_ error codes if the need arose in future. A small refactor would also see it more widely available (perhaps even via `State`) if/when the need arises. I can't see a reason to do it today, so I haven't.

I'm going to attempt an optimisation to remove the `HashSet` (and therefore the memory allocation) from `RouteNonMatch`, but this is otherwise ready for review / consideration.

### DIFF
--- a/gotham/src/router/tree/node.rs
+++ b/gotham/src/router/tree/node.rs
@@ -134,38 +134,31 @@ impl Node {
         &'a self,
         state: &State,
     ) -> Result<&'a Box<Route + Send + Sync>, RouteNonMatch> {
-        enum Selection<'b> {
-            Init,
-            Match(&'b Box<Route + Send + Sync>),
-            NonMatch(RouteNonMatch),
+        let mut err: Result<(), RouteNonMatch> = Ok(());
+
+        for r in self.routes.iter() {
+            match r.is_match(state) {
+                Ok(()) => {
+                    trace!("[{}] found matching route", request_id(state));
+                    return Ok(r);
+                }
+                Err(e) => match err {
+                    Err(e0) => err = Err(e.union(e0)),
+                    Ok(()) => err = Err(e),
+                },
+            }
         }
 
-        let iter = self.routes.iter();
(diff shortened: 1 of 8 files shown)


gotham/src/router/tree/node.rs
```diff
@@ -129,24 +130,47 @@ impl Node {
...
+        state: &State,
+    ) -> Result<&'a Box<Route + Send + Sync>, RouteNonMatch> {
+        enum Selection<'b> {
+            Init,
+            Match(&'b Box<Route + Send + Sync>),
+            NonMatch(RouteNonMatch),
+        }
+
+        let iter = self.routes.iter();
+        let selection = iter.fold(Selection::Init, |s, r| match s {
```
bradleybeddoes: Thoughts on making this a `for` loop (with mutable accumulator for `Selection::NonMatch`) so we can short circuit on a `Selection::Match` result?

bradleybeddoes: Alternatively it seems like IterTools has a `fold_while` function, which is basically the same thing, https://docs.rs/itertools/*/itertools/trait.Itertools.html#method.fold_while but I'm not sure on adding another dependency just for that.

bradleybeddoes: @smangelsdorf did you have any thoughts here? Figured this comment slipped through the cracks.

smangelsdorf: Sorted. It had slipped under a pile of other work, though I was always going to come back to it (eventually 😛).

Issue #95, opened 2018-01-14
Question: start gotham server using a provided Handle


kierdavis:
I would like to run two tokio-based servers in parallel in my application, one of which is a Gotham HTTP server. As far as my understanding of tokio goes, it ought to be possible to create one event loop and register both services onto it.

Is there a way to start a Gotham server using a provided tokio `Handle`, rather than letting it create its own event loop? If not, and you feel that its a reasonable feature to implement, I'm happy to work on adding this behaviour.

ChristophWurst:
Yes, that's possible. See https://github.com/gotham-rs/gotham/pull/66#issuecomment-344761770

kierdavis:
Great, thanks! I'll check that out.

bradleybeddoes:
Hi @kierdavis you absolutely can and @ChristophWurst has pointed out one option for doing that which should work for `0.1.x`.

However we've recently merged in https://github.com/gotham-rs/gotham/pull/89 which may have broken this method if you're using Gotham master (which represents the 0.2 release, I'd actually recommend using this if you're embarking on a new application, there are a lot of bug fixes and new features coming in 0.2)

What would be really useful, if you have the time, is a `start_with_handler` function that is essentially the same as https://github.com/gotham-rs/gotham/blob/master/gotham/src/lib.rs#L61 but does not create a Tokio core when you into the actual `serve` function, as it does now https://github.com/gotham-rs/gotham/blob/master/gotham/src/os/unix.rs#L46

kierdavis:
Thanks for the info! I'm currently running v0.1 as that the latest available on crates.io, but I'm all for switching to 0.2 when it becomes available.

I'm happy to have a crack at implementing `start_with_handler` as you say. Looking at the sources it seems it should be trivial to pass a `Handle` down to where it's needed. One question I have regarding this is how to handle multithreading in this scenario - currently we create one event loop per thread, so would we want to accept a list of `Handles` or share one between all threads?

As my original question has been answered (@ChristophWurst's suggestion is working perfectly for me) I'll rename this issue to be more inline with the feature request side of things.

bradleybeddoes:
So @smangelsdorf and I have just thrown this very discussion around ourselves.

What `start` does nicely now is ensuring that there is a Tokio Core per thread and we create n == CPU core threads at startup.

[...]

PR #121, opened 2018-01-24
Refactor core creation out of serve & listen

kierdavis:
This change decouples the code that creates a `Core` and `Handle` from the code that uses it - the former is pulled out into new functions `start_code` and co, while the latter remains in `serve` and `listen`.

As discussed in #95, the intention is for `serve`, `listen` and other necessary parts of the code can then be made public, providing an interface by which a user can start a gotham server on a provided `Handle` instance.

### DIFF
--- a/gotham/src/os/unix.rs
+++ b/gotham/src/os/unix.rs
@@ -1,10 +1,11 @@
+use std::io;
 use std::net::{SocketAddr, TcpListener, ToSocketAddrs};
 use std::thread;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


smangelsdorf:
Thanks for this. I'm happy with this as a refactor, but there will need to be some more consideration about how to expose the `serve` function before it is made `pub`. I'll update #95 with more of my thoughts on how we could proceed from here.

kierdavis:
No worries!

PR #125, opened 2018-02-01
Switch to serde for path and query string extractors

smangelsdorf:
This reimplements our `PathExtractor` and `QueryStringExtractor` ideas as a serde `Deserializer`, so that we populate app-provided structs via a derived `Deserialize` implementation. Aside from a few things moving to different modules, the primary difference in how this is used is:

**Old:**

```rust
#[derive(PathExtractor, StateData, StaticResponseExtender)]
struct MyPathParams {
    id: u32,
}
```

**New:**

```rust
#[derive(Deserialize, StateData, StaticResponseExtender)]
struct MyPathParams {
    id: u32,
}
```

`PathExtractor` / `QueryStringExtractor` types are still applied to routes in exactly the same way:

```rust
build_simple_router(|route| {
    route.get("/:id")
        .with_path_extractor::<MyPathParams>()
        .to(my_handler);
})
```

The advantages of this approach are that apps can make use of **any** type which can be deserialized with serde, and we get all the extra features of serde for free (e.g. `#[serde(rename = "..")]`, `#[serde(deserialize_with = "..")]`).

I'd appreciate feedback from anyone and everyone who has time to review this. There's a fair bit of code here, but I've tried to go fairly heavy on the code comments and tests just to prove it works in enough base cases that serde can handle the edge cases for us.


gotham/src/extractor/internal.rs
```diff
@@ -0,0 +1,1150 @@
...
+        );
+    };
+}
+
+/// This trait represents the possible types that we can deserialize from when we're using
+/// extractors. The concrete values of this are all `IteratorAdaptor` types, and this trait is
+/// primarily giving us one place to deal with the type structure and expose the `next` function
+/// that we require.
+///
+/// See `from_segment_mapping` and TODO: `from_query_string` for how the values are constructed.
```
bradleybeddoes: minor nit: there are a few of these TODO scattered about, these appear to be resolved now, should they be cleaned up?

smangelsdorf: Done.

bradleybeddoes:
@smangelsdorf this is now conflicted with master

smangelsdorf:
Classic me, being in conflict with myself.

Issue #77, opened 2017-12-26
Add example that shows usage of Diesel middleware


bradleybeddoes:

n-pochet:
Hi,
I discovered recently your project and I think it is really nice.
I really would like to help. 
Because for one of my project I would need to use the Diesel middleware, I would like to help on this issue.
For the moment, I'm still in the process of learning Gotham and trying to understand how to use all the components (especially middlewares and pipelines).
For the example app, I thought about making a simple app that would be able to retrieve some fields in the DB in reaction to a GET request. That app would serve some posts (like in the Getting started app of Diesel [Diesel Getting Started](http://diesel.rs/guides/getting-started/)) but using SQlite instead of Postgres.
Ideally, it would be nice to have all the CRUD operations implemented.
What do you think?
Any advice regarding the code organization? The basic example of Diesel provides some modules and a main `lib.rs` file.
Thank you beforehand for the help!
Kind regards,

Nicolas

bradleybeddoes:
Hi @n-pochet welcome to the Gotham project and thank you for your kind words.

I see you've submitted PR #126 which is awesome, I'll make some comments and work with you on this over there, might be a day or so until I get to do that, very busy weekend on this side 👍.

msrd0:
I believe this has been done in #313

Issue #108, opened 2018-01-23
Create Path Extraction example


bradleybeddoes:

n-pochet:
Hi, since I saw yesterday's PR #125 merge, I would like to try to make an example of it.
I thought that at least something like this would be interesting:
```
GET localhost/myname
...

Hello, myname!
```
It is really straightforward to implement.
@bradleybeddoes Did you think to something else?

bradleybeddoes:
Sounds good, perhaps `basic_path_extractor`.

What I would like to do is try and unify the examples around what is currently shown in https://github.com/gotham-rs/gotham/blob/master/examples/basic_router/src/main.rs in terms of theme, that being a web store, which is hopefully a generic enough concept to assist folks in their learning of Gotham.

In this case how would you feel about the request being for `/widgets/:name`? (No need to replicate all routes shown in `basic_router`). The result body being something like `format!("Product: {}", name)`. 

The ultimate goal will be creation of a final example (at some point in the future), that combines all the concepts shown in all the previous examples, potentially in more advanced ways, leading to a basic, but functional, web store.

PR #126, opened 2018-02-01
Diesel example

n-pochet:
This PR addresses the issue #77 
It is the first draft and could certainly be improved.
Thank you beforehand for your advice and remarks!


Cargo.toml
```diff
@@ -11,4 +11,5 @@ members = [
     "examples/hello_world",
     "examples/hello_router",
     "examples/basic_router",
+    "examples/basic_diesel",
```
bradleybeddoes: Suggest we call this example simply "diesel"

examples/basic_diesel/.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=.posts.db
```
bradleybeddoes: Suggest we hardcode this for the example, goal being to streamline 3rd party overheads/concepts as much as possible.

One of the examples I have planned is going to give some ideas for external configuration loading so we'll cover this then.

examples/basic_diesel/migrations/2018-01-31-183719_create_posts/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
```
bradleybeddoes: suggest remove comment

examples/basic_diesel/migrations/2018-01-31-183719_create_posts/up.sql
```diff
@@ -0,0 +1,7 @@
+-- Your SQL goes here
```
bradleybeddoes: suggest removing comment

examples/basic_diesel/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+use diesel::sqlite::SqliteConnection;
+use r2d2_diesel::ConnectionManager;
+use r2d2::{Pool, PooledConnection};
+use futures::{future, Future, Stream};
+use std::str;
+use basic_diesel::models::NewPost;
+
+
+// The URL of the database.
+static DATABASE_URL: &'static str = "posts.db";
```
bradleybeddoes: I note the database was committed, suggest removing the binary, it won't be necessary for folks.

examples/basic_diesel/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+            create_table(conn);
+            diesel::insert_into(posts::table)
+                .values(&new_post)
+                .execute(conn)
+        }
+    }
+}
+
+/// Create the table if needed.
+fn create_table(conn: &SqliteConnection) {
```
bradleybeddoes: checkout `embed_migrations!` as part of https://docs.rs/diesel_migrations/1.1.0/diesel_migrations/macro.embed_migrations.html which will help here and cleanup the Err case at https://github.com/gotham-rs/gotham/pull/126/files#diff-92b023737c440588ef77ba19c716ee1eR54

n-pochet:
@bradleybeddoes Thanks for the review. I'll make the changes taking your advice into account.

Cargo.toml
n-pochet: Indeed. But I still need to name my package in `Cargo.toml` `basic_diesel` or I will get a conflict with the real `diesel` crate.

bradleybeddoes:
I'm just thinking out loud here and feel free to not like this idea, but following my comment on #108 I started considering this PR again.

I was wondering what you thought of also giving this example the 'web store' treatment. Essentially changing it from creating and listing `Post` to creating and listing `Widget` as part of the web store "api" that I started to hint at here: https://github.com/gotham-rs/gotham/blob/master/examples/basic_router/src/main.rs#L62

n-pochet:
Sure. It makes more sense to have a "unified" set of examples. I'll change that.
Do you have any idea of what the definition of a `Widget` should be?

bradleybeddoes:
Oops! I guess this could be very easily confused by folks who haven't seen that before!!.

Perhaps a more universal name here would be `Product`/`products` instead of `Widget`/`widgets`. 

Maybe the products we "sell" could be some of the Rust language merchandise I've seen floating about such as stickers, shirts and the like. e.g:

* https://www.redbubble.com/shop/rust+language
* https://devswag.com/search?q=Rust

I'm open to suggestions!

nyarly:
@n-pochet It's been some time since this PR saw attention (from us!) How do you feel about it these days? Should it get a new review pass?

n-pochet:
Hi @nyarly,

Sorry for the late reply. I have a ton of other projects running.
I rebased and fixed the issues in the example. Feel free to review.

bradleybeddoes:
Closing due to lack of input, please re-open if still a wanted PR.

PR #127, opened 2018-02-05
Add Hello Header example

eguven:
Example setting custom headers.

Fixes https://github.com/gotham-rs/gotham/issues/117. I'm very new to Rust so please do point out if there are more idiomatic ways of doing this or if I need to extend the example further.


bradleybeddoes:
G'day @eguven welcome to the Gotham community (and Rust!), thanks for your contribution here 👍.

I'm a day or so behind reviews but just a quick note here that it would be great if your example could use a `router` function similar to [this one](https://github.com/gotham-rs/gotham/blob/master/examples/hello_router/src/main.rs#L26). 

I believe most folks will be looking to work with Gotham via the `Router` so having this appear in all of our examples (other than `hello_world` naturally!!) seems like a good pattern to carry throughout the crates we're shipping in the `examples` directory.

eguven:
I added the `Router` change and moved the example to after `hello_router` in readme.

examples/hello_header/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "hello_header"
+version = "0.1.0"
+authors = ["Bradley Beddoes <bradleybeddoes@gmail.com>",
+           "Shaun Mangelsdorf <s.mangelsdorf@gmail.com>"]
```
bradleybeddoes: Don't forget to add your name here!

bradleybeddoes:
Hi @eguven just wondering if this was passed through rustfmt? (0.3.4-nightly or later).

We're going to look at making this a requirement of the Gotham repository CI process before 0.2 for all submissions so it would be great to start now!.

I know you said you're new to Rust so if you haven't come across rustfmt yet here are the details https://github.com/rust-lang-nursery/rustfmt.

eguven:
Yea, it turns out I had the deprecated rustfmt 0.10.0.

    $ rustup run nightly rustfmt -V
    0.3.6-nightly (e0e3e22 2018-01-18)

examples/hello_header/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+
+// this macro defines our custom header
+header! { (GothamHeader, "X-Gotham") => [String] }
+
+/// Create a `Handler` that adds a custom header.
+pub fn say_hello_through_header(state: State) -> (State, Response) {
+    let mut res = create_response(
+        &state,
+        StatusCode::Ok,
+        Some((b"".to_vec(), mime::TEXT_PLAIN)),
```
bradleybeddoes: oops I missed this earlier, all this needs to be is a `None` as you don't want/need a body + mime type in this response.

PR #132, opened 2018-02-07
Support Cookie Prefixes

colinbankier:
This is one possible implementation for #6 
This provides a helper to ensure the associated attributes are set when a cookie name is set with a prefix. This was generally inspired by the way Play framework does this.

If attributes are not set in line with the prefix, they are changed when the cookie is serialised. A warning is emitted - to disable the warning a user would either have to set the cookie attributes correctly or remove the prefix.

See cookie prefix spec: https://tools.ietf.org/html/draft-west-cookie-prefixes-05

### DIFF
--- a/gotham/src/middleware/session/mod.rs
+++ b/gotham/src/middleware/session/mod.rs
@@ -26,6 +26,9 @@ mod rng;
 pub use self::backend::{Backend, NewBackend};
 pub use self::backend::memory::MemoryBackend;
 
+const SECURE_COOKIE_PREFIX: &'static str = "__Secure-";
+const HOST_COOKIE_PREFIX: &'static str = "__Host-";
+
 /// Represents the session identifier which is held in the user agent's session cookie.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct SessionIdentifier {
@@ -64,7 +67,7 @@ enum SameSiteEnforcement {
 /// By default, the cookie has the name "_gotham_session", and the cookie header includes the
 /// `secure` flag.  `NewSessionMiddleware` provides functions for adjusting the
 /// `SessionCookieConfig`.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct SessionCookieConfig {
     // If `Expires` / `Max-Age` are ever added update `reset_session` to allow for them.
     name: String,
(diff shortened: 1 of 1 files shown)


gotham/src/middleware/session/mod.rs
```diff
@@ -91,36 +94,74 @@ impl Default for SessionCookieConfig {
...
         }
 
         cookie_value.push_str("; Path=");
-        cookie_value.push_str(&self.path);
+        cookie_value.push_str(&cookie_config.path);
 
         cookie_value
     }
+
+    fn validate_prefix(&self) -> SessionCookieConfig {
```
colinbankier: I realised this could return an `Option<SessionCookieConfig>` instead, so that in the case that it doesn't change it doesn't need to be copied. Is that a worthwhile change?

smangelsdorf: (Avoiding the need to `.clone()` but also avoiding the need for an `Option<_>` return type)

colinbankier: I'll try it out  - it also crossed my mind, but because it's called from `to_cookie_string` which takes '&self` I didn't think it was possible. I may learn new things about ownership though...

gotham/src/middleware/session/mod.rs
```diff
@@ -93,34 +96,84 @@ impl SessionCookieConfig {
         // Ensure this is always enough to prevent re-allocs
         let mut cookie_value = String::with_capacity(255);
 
-        cookie_value.push_str(&self.name);
+        let validated_config = self.validate_prefix();
+        let cookie_config = if let Some(ref updated_config) = validated_config {
```
colinbankier: I added a little dance to avoid copying the cookie config if it doesn't need to be updated.
I don't think `validate_prefix` can take an owned `self` to avoid the `Option` since `to_cookie_string` only has a borrowed ref. 
If there's a nicer way to do this, happy to learn :)

smangelsdorf: I understand. Sorry, that's my mistake for making the suggestion without considering the change in full 😁 

Would it be possible to do this logic up-front when the middleware is being built, rather than processing it per-request? At that point, all the builder methods take an owned `self`, but copying is cheap anyway since it's a one-time setup when booting the app.

The downside is that there wouldn't be a warning for every request, but a positive is that we could refuse to set options that disagree with the cookie prefix.

(When I made my prior suggestion, I was incorrectly assuming this is how it worked.)

colinbankier: Yeah, I agree - it looks like it could be done once in `rebuild_new_session_middleware` (rather than checking in each `with_cookie_*` method). I'll move it there :smile:

gotham/src/middleware/session/mod.rs
```diff
@@ -491,7 +541,7 @@ where
         cookie_config: SessionCookieConfig,
     ) -> NewSessionMiddleware<B, T> {
         NewSessionMiddleware {
-            cookie_config: Arc::new(cookie_config),
+            cookie_config: Arc::new(cookie_config.validate_prefix()),
```
colinbankier: Revised - validating the prefix when `NewSessionMiddleware` is built rather than on cookie serialisation.

PR #133, opened 2018-02-07
Add Query String Extractor example

n-pochet:
Create example and update README.md and Cargo.toml
This PR addresses issue #109 
It is possible, when running the example to:
```
curl -v 'http://localhost:7878/widgets?name=t-shirt'
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 7878 (#0)
> GET /widgets?name=t-shirt HTTP/1.1
> Host: localhost:7878
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 31
< Content-Type: text/plain
< X-Request-ID: 09542fc7-6739-465c-9ee8-8aa18c3a5271
< X-Frame-Options: DENY
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Runtime-Microseconds: 308
< Date: Wed, 07 Feb 2018 18:52:15 GMT
< 
* Connection #0 to host localhost left intact
{"name":"t-shirt","price":15.5}%
```


examples/basic_query_extractor/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+    gotham::start(addr, router())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use gotham::test::TestServer;
+
+    #[test]
+    fn index_not_found() {
```
bradleybeddoes: I'm fine to remove this test so we're focused on tests specifically related to this example only, not the Router as a whole (which is tested elsewhere!!).

examples/basic_query_extractor/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+    name: String,
+}
+
+/// `Product` struct
+///
+/// It represents products that can be queried
+#[derive(Serialize)]
+struct Product {
+    name: String,
+    price: f32,
```
bradleybeddoes: How about we switch `price` for `description` for now? 

I wouldn't want to give the wrong impression about types that are considered useful in representing currency.

examples/basic_query_extractor/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+        None => return (state, Response::new().with_status(StatusCode::NotFound)),
+    };
+    // Create a response using `Product` serialized to JSON
+    let res = create_response(
+        &state,
+        StatusCode::Ok,
+        Some((
+            format!("{}", serde_json::to_string(&product).unwrap())
+                .into_bytes(),
+            mime::TEXT_PLAIN,
```
bradleybeddoes: JSON mime type

examples/basic_query_extractor/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+        price: 15.5,
+    });
+    products.push(Product {
+        name: "sticker".to_string(),
+        price: 1.5,
+    });
+    products.push(Product {
+        name: "mug".to_string(),
+        price: 10.0,
+    });
```
bradleybeddoes: I'd be inclined to just add the one product here for brevity in the example

examples/basic_query_extractor/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+    // Generate the products and try to find a match
+    let products = generate_products()
+        .into_iter()
+        .filter(|p| p.name == name)
+        .collect::<Vec<Product>>();
+    // Check if a product is found. If not, return `StatusCode::NotFound`
+    let product = match products.get(0) {
+        Some(p) => p,
+        None => return (state, Response::new().with_status(StatusCode::NotFound)),
+    };
```
bradleybeddoes: We can simplify a number of lines of this code to make it easier for folks to read.

I wrote up an example in Rust playground which I am hoping is useful: https://play.rust-lang.org/?gist=b2234b1032b2c1882cfd42eaeb513cb4&version=stable

examples/basic_query_extractor/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+        name: "mug".to_string(),
+        price: 10.0,
+    });
+    products
+}
+
+/// Function to handle the `GET` requests coming to `/widgets/:name`
+fn get_product_handler(mut state: State) -> (State, Response) {
+    // Extract `name` from `state`
+    let name = state.take::<QueryParam>().name;
```
bradleybeddoes: I feel like highlighting the [FromState](https://github.com/gotham-rs/gotham/blob/935ab297113918e050064eab1addcadc619e6ed9/gotham/src/state/from_state.rs) trait in our examples is a good idea, thus this would become:

```
let query_param = QueryParam::borrow_from(&state); 
```

n-pochet:
I did a mistake while rebasing. I will reopen it.

PR #150, opened 2018-02-20
Docs review A-H

smangelsdorf:
Broken down by top-level module name, I've gone as far as `gotham::http` with this PR.

Follows from #149 to avoid mass conflicts there, but the [diff can also be viewed separately](https://github.com/gotham-rs/gotham/compare/visibility-review...docs-review-1).

### DIFF
--- a/gotham/src/extractor/mod.rs
+++ b/gotham/src/extractor/mod.rs
@@ -1,8 +1,13 @@
 //! Extracts request data into type-safe structs using Serde.
 //!
-//! The request data is extracted by the `Route` implementation when dispatching the request. An
-//! application-provided data structure is used to deserialize the data and store it within the
-//! request `State` before the request is dispatched to the `Handler`.
+//! Extractors are added to route definitions when defining a `Router`. The `PathExtractor` and
+//! `QueryStringExtractor` traits provide usage examples. For further examples, see the Gotham
+//! Router examples.
+//!
+//! The request data is extracted by the `Route` implementation when dispatching the request. The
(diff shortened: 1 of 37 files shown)


gotham/src/lib.rs
```diff
@@ -1,6 +1,6 @@
-//! Gotham - A flexible web framework that does not sacrifice safety, security or speed.
+//! Gotham &ndash; A flexible web framework that does not sacrifice safety, security or speed.
 //!
-//! You can find out more about Gotham, including where to get help,  at https://gotham.rs.
+//! You can find out more about Gotham, including where to get help, at <https://gotham.rs>.
```
bradleybeddoes: Phrase itself has since changed to:

"A flexible web framework that promotes stability, safety, security and speed."

gotham/src/extractor/mod.rs
```diff
@@ -1,8 +1,13 @@
 //! Extracts request data into type-safe structs using Serde.
 //!
-//! The request data is extracted by the `Route` implementation when dispatching the request. An
-//! application-provided data structure is used to deserialize the data and store it within the
-//! request `State` before the request is dispatched to the `Handler`.
+//! Extractors are added to route definitions when defining a `Router`. The `PathExtractor` and
+//! `QueryStringExtractor` traits provide usage examples. For further examples, see the Gotham
```
bradleybeddoes: Path and QueryString are actually standalone functionality groups now. Perhaps we just dump the "For further examples" sentence entirely.

PR #168, opened 2018-03-01
Add examples using sessions

illicitonion:
One is for a basic data type (usize), and the other shows using your own
custom data type.


bradleybeddoes:
Thanks for the PR!.

I've not had a chance to go through it in detail yet (I'll likely need a day or two on that one) but on the surface I'm thinking that these are possibly more suited as "session" examples (a new functionality category/directory under examples) rather than "cookie" examples. What they really do is help folks make use of Gotham session support.

I feel that "cookie" examples are probably the sort of thing that would show folks how to access and set cookies and use some of the cookie related API we have in Gotham.

Any thoughts?

Also PR failed CI due to `rustfmt` diffs. Not a big deal, just needs to have `rustfmt` version `0.3.8` run against the proposed code here.

Thanks again!

illicitonion:
> I'm thinking that these are possibly more suited as "session" examples (a new functionality category/directory under examples) rather than "cookie" examples. What they really do is help folks make use of Gotham session support.

Yeah, that's totally fair :) I've renamed them accordingly.

Does gotham actually have any more featured support for cookies than giving raw header access from hyper? I couldn't see anything, which is why I assumed this was how cookies were generally intended to be used. I'm happy to put together a separate example PR using cookies too, if you can point me at how they're intended to be used...

> Also PR failed CI due to rustfmt diffs. Not a big deal, just needs to have rustfmt version 0.3.8 run against the proposed code here.

Oh dear - it looks like I had a different version of rustfmt installed. A few cycles of removing and updating assorted nightly rust versions, and fmt versions, has got me able to reproduce travis's results - fixed :)

examples/sessions/custom_data_type/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+    last_visit: String,
+}
+
+/// Handler function for `GET` requests directed to `/`
+///
+/// Each request made will update state about your recent visits, and report it back.
+fn get_handler(mut state: State) -> (State, Response) {
+    let maybe_visit_data = {
+        let visit_data: &Option<VisitData> = SessionData::<Option<VisitData>>::borrow_from(&state);
+        visit_data.clone()
```
alsuren: I was trying to work out whether this could be done without the clone() and without the type-coercing noise. I think that deref() and deref_mut() help a lot here. This is what I came up with: 

```
fn get_handler(mut state: State) -> (State, Response) {
    let (visit_count, body) = match SessionData::<Option<VisitData>>::borrow_from(&state).deref() {
        &Some(ref visit_data) => (
            visit_data.count,
            format!(
                "You have visited this page {} time(s) before. Your last visit was {}.\n",
                visit_data.count, visit_data.last_visit,
            ),
        ),
        &None => (0, "You have never visited this page before.\n".to_owned()),
    };
    ...
    {
        *SessionData::<Option<VisitData>>::borrow_mut_from(&mut state).deref_mut() =
            Some(VisitData {
                count: visit_count + 1,
                last_visit: format!("{}", time::now().rfc3339()),
            });
    }
    (state, res)
}
```

I'm not sure if I like my match that returns a tuple of (visit_count, body) or not: I think my mind is still in python mode. I suspect that it would get increasingly awkward as the view gets more complicated. Might be better to stick with .clone().

bradleybeddoes:
> Does gotham actually have any more featured support for cookies than giving raw header access from hyper? I couldn't see anything, which is why I assumed this was how cookies were generally intended to be used.

Right now there isn't however some of the cookie code we have in place for sessions may be extractable as generic helper methods for working with cookies, much like `create_response` is for working with responses.

So for now any examples we create for cookies would be just showing usage of the underpinning hyper API but there is some value in that for newcomers who are still finding their way around.

Issue #115, opened 2018-01-23
Create async handler example


bradleybeddoes:

thedodd:
A super easy way to do it (based on some GraphQL work I've been doing):
```rust
// ... snip ...

    /// The handler for executing GraphQL requests.
    pub fn graph(state: State) -> Box<HandlerFuture> {
        Box::new(GatewayService::_graph(state))
    }

    /// Business logic implementation of handling GraphQL requests.
    #[async]
    fn _graph(mut state: State) -> Result<(State, Response), (State, HandlerError)> {
        // Begin tracing this graph request.
        let uri = String::from(Uri::borrow_from(&state).path());
        let _span = (*APISpan::borrow_from(&state)).child(uri, |ctx| ctx.start());

        // Extract the request body.
        let body_res = await!(Body::take_from(&mut state).concat2().map(|chunks| chunks.to_vec()));
        // ... snip ...
    }

// ... snip ...
```

Works like a charm; though I would love to be able to not have the top-level indirect for boxing the future (having just one fn, instead of `graph` & `_graph`).

alsuren:
I'm currently in the process of writing some examples that use hyper::Client and `.and_then()` and `.join()`.

@thedodd does `#[async]` still require nightly? I will probably mention it, but I'm trying to stick to things that are available in stable rust.

bradleybeddoes:
Great async handler examples created by @alsuren available at https://github.com/gotham-rs/gotham/tree/master/examples/handlers

PR #170, opened 2018-03-04
Add more involved async handler examples

alsuren:
Three examples of different ways to skin the same cat (in this case, making http requests inside a Handler).

I give 3 approaches (using `Future.and_then()`, using `Stream.fold()`, and using `Future.join()`) and discuss the merits of each.

I don't give any examples of `#[async]` handlers (as suggested in https://github.com/gotham-rs/gotham/issues/115#issuecomment-370201791 ) but I do give links to futures-await for anyone that's using nightly.

I should also say thank you to @alexreg for pairing with me on this. My first shot at this was ugly as sin, and now it feels a bit more like idiomatic Rust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ members = [
 
     # handlers
     "examples/handlers/request_data",
+    "examples/handlers/async_handlers",
 
     # example_contribution_template
     "examples/example_contribution_template/name"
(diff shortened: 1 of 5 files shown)


smangelsdorf:
Thanks, this looks like a great way to present async handlers that do something more complex than just reading the request body. I've kicked off Travis CI again to resolve the timeout it encountered.

I'll have a look at this in more depth when I get some time.

PR #174, opened 2018-03-05
Only install rustfmt once

illicitonion:
This regresses us to a more official release (0.3.8 rather than 0.4.0)
but saves 8~10 minutes of compiling rustfmt from scratch.


illicitonion:
Ironically, this vaguely obsoletes https://github.com/gotham-rs/gotham/pull/173 which I'm happy to revert if you'd like (though having a separate shard isn't a bad thing either way :))

smangelsdorf:
I'm happy to keep the separate rustfmt shard, personally.

PR #176, opened 2018-03-07
Add example using cookies

illicitonion:
This is just using the raw hyper Cookie/SetCookie types, as that's
currently the public interface offered by Gotham. Perhaps in the future
we will offer a more typesafe API, and some better defaults.


bradleybeddoes:
Thanks very much for this one @illicitonion - If you do have any thoughts for a better API we could put in place for Gotham around cookies I'd be keen to talk about it.

PR #177, opened 2018-03-09
Allow Router builder to use custom RouteMachers

bradleybeddoes:
This support is necessary for folks who want to dispatch a request on more than matching the method used (or potentially not concern themselves with the method used at all).

A subsequent PR will introduce an `add_matcher` so something like:

```
route.get("/").add_matcher(accept_matcher).to(handler)
```

Will also be feasible.

### DIFF
--- a/gotham/src/router/builder/draw.rs
+++ b/gotham/src/router/builder/draw.rs
@@ -11,6 +11,7 @@ use router::builder::{AssociatedRouteBuilder, DelegateRouteBuilder, RouterBuilde
                       SingleRouteBuilder};
 use router::tree::node::{NodeBuilder, SegmentType};
 use router::tree::regex::ConstrainedSegmentRegex;
+use router::route::matcher::RouteMatcher;
 
 /// The default type returned when building a single route. See
 /// `router::builder::DefineSingleRoute` for an overview of the ways that a route can be specified.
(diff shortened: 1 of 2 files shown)


gotham/src/router/builder/draw.rs
```diff
@@ -393,11 +399,88 @@ where
...
+    /// #       .perform()
+    /// #       .unwrap();
+    /// #   assert_eq!(response.status(), StatusCode::Accepted);
+    /// # }
+    /// ```
+    fn explicit_request<'b, M>(
+        &'b mut self,
+        matcher: M,
+        path: &str,
+    ) -> ExplicitSingleRouteBuilder<'b, M, C, P>
```
smangelsdorf: This started somewhat as an objection to the word "explicit" here, but when I couldn't think of a better word I turned my eye to the API itself.

What if we instead refactored [the existing `request` method](https://docs.rs/gotham/0.2.0/gotham/router/builder/trait.DrawRoutes.html#method.request) to fill this job, by doing something like:

```rust
trait IntoRouteMatcher {
    type Output: RouteMatcher;
    fn into_route_matcher(self) -> Self::Output;
}

impl IntoRouteMatcher for Vec<Method> { .. }
impl<M> IntoRouteMatcher for M where M: RouteMatcher { .. }
```

```rust
fn request<'b, M>(
    &'b mut self,
    m: M,
    path: &str,
) -> ExplicitSingleRouteBuilder<'b, M::Output, C, P>
where
    M: IntoRouteMatcher,
{
    ..
}
```

bradleybeddoes: This will make it much cleaner, i'll get this implemented.

bradleybeddoes:
back to you @smangelsdorf

PR #185, opened 2018-03-18
Correctly specify minimum dependency versions

### LINKED ISSUE #180: no method named `serve_connection` found for type `hyper::server::Http` in the current scope
I updated to Gotham 0.2.0 and got this error:

```rust
error[E0599]: no method named `serve_connection` found for type `hyper::server::Http` in the current scope
   --> /home/judson/.cargo/registry/src/github.com-1ecc6299db9ec823/gotham-0.2.0/src/test/mod.rs:172:14
    |
172 |             .serve_connection(ss, service)
    |             
```

I suspect that I need to update hyper, which suggests that Gotham doesn't report its dependency adequately?

smangelsdorf:
There are cases where Gotham (and by extension, Gotham apps) won't build correctly with the currently specified minimum version. Specifying the minimum version which permits Gotham to build correctly ensures that it can't be broken by apps with old `Cargo.lock` files.

Fixes #180

This also brings the Travis CI changes from #173 and #174 into the 0.2-maint branch.

### DIFF
--- a/gotham/Cargo.toml
+++ b/gotham/Cargo.toml
@@ -12,14 +12,14 @@ categories = ["web-programming::http-server"]
 keywords = ["http", "async", "web", "framework", "blockchain"]
 
 [dependencies]
-log = "0.4"
-hyper = { version = "0.11", features = [] }
+log = "0.4.1"
+hyper = { version = "0.11.7", features = [] }
 serde = "1"
 serde_derive = "1"
 bincode = "1"
 mime = "0.3"
 futures = "0.1"
-tokio-core = "0.1"
+tokio-core = "0.1.13"
 mio = "0.6"
 borrow-bag = "1"
 url = "1"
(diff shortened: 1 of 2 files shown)


gotham/Cargo.toml
```diff
@@ -12,14 +12,14 @@ categories = ["web-programming::http-server"]
 keywords = ["http", "async", "web", "framework", "blockchain"]
 
 [dependencies]
-log = "0.4"
-hyper = { version = "0.11", features = [] }
+log = "0.4.1"
+hyper = { version = "0.11.7", features = [] }
```
bradleybeddoes: epic nit, ignorable.

Given we're not using any hyper features should we take the opportunity to revert back to `hyper = "0.11.7"` format?

smangelsdorf: I thought the goal here was to explicitly opt out of hyper's default features. From looking at the latest hyper, that seems to opt us out of tokio-proto compatibility (since we don't need it). I might be misreading though.

bradleybeddoes: 👍 carrying on.

Issue #190, opened 2018-03-24
Add better rustdoc example code for `with_pipeline_chain` in the router builder


smangelsdorf:
The current example only shows how to switch to using _no_ pipelines, but not how to use a different set of pipelines. We need a second doc example showing how to use multiple pipelines for a subset of routes (that's the whole point of pipelines, after all).

Something I knocked together quickly but didn't test:

https://gist.github.com/smangelsdorf/eb0d15dbad2697b3ad1f5808bf114993

ParisLiakos:
@smangelsdorf this can be closed now, right?

smangelsdorf:
It can. Good catch 👍

PR #191, opened 2018-03-24
Expand the doc example of DrawRoutes::with_pipeline_chain()

ParisLiakos:
I am not sure whether adding a second Session middleware makes sense, but i think its fine for an example :)

Many thanks to @smangelsdorf for his gist in #190!


smangelsdorf:
Noting that this passes stable/nightly but rustfmt is complaining (#193 is one possible fix, but it wasn't caused by this PR.)

PR #192, opened 2018-03-24
Redirect sugar

onelson:
The type for the location string is made to match the signature of
hyper's Location header. Seems odd to me to reach to `Cow` for this, but
surely the hyper folks know something I don't.

Follow up rev with the old `cargo fmt` on it.

### DIFF
--- a/gotham/src/http/response/mod.rs
+++ b/gotham/src/http/response/mod.rs
@@ -70,7 +70,6 @@ pub fn create_response(state: &State, status: StatusCode, body: Option<Body>) ->
     res
 }
 
-
 /// Produces a simple empty `Response` with a `Location` header.
 ///
 /// This function is intended as a convenience. For more elaborate redirection
@@ -133,15 +132,18 @@ pub fn create_response(state: &State, status: StatusCode, body: Option<Body>) ->
 /// #     }
 /// # }
 /// ```
-pub fn create_simple_redirect<L: Into<Cow<'static, str>>>(state: &State, permanent: bool, location: L) -> Response {
-    let status = if permanent { StatusCode::PermanentRedirect } else { StatusCode::TemporaryRedirect };
+pub fn create_simple_redirect<L: Into<Cow<'static, str>>>(
+    state: &State,
+    permanent: bool,
+    location: L,
+) -> Response {
+    let status = if permanent {
+        StatusCode::PermanentRedirect
+    } else {
+        StatusCode::TemporaryRedirect
+    };
     let mut res = Response::new().with_status(status);
-    set_headers(
-        &state,
-        &mut res,
-        None,
-        None,
-    );
+    set_headers(&state, &mut res, None, None);
     res.headers_mut().set(Location::new(location));
     res
 }
--- a/gotham/src/http/response/mod.rs
+++ b/gotham/src/http/response/mod.rs
@@ -1,7 +1,8 @@
 //! Helpers for HTTP response generation
 
+use std::borrow::Cow;
 use hyper::{Method, Response, StatusCode};
-use hyper::header::{ContentLength, ContentType};
+use hyper::header::{ContentLength, ContentType, Location};
 use mime::Mime;
 
 use state::{request_id, FromState, State};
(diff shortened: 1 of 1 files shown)


gotham/src/http/response/mod.rs
```diff
@@ -69,6 +70,84 @@ pub fn create_response(state: &State, status: StatusCode, body: Option<Body>) ->
...
+/// #
+/// #         assert_eq!(response.status(), StatusCode::TemporaryRedirect);
+/// #         assert_eq!(
+/// #             response.headers().get::<Location>(),
+/// #             Some(&Location::new("/detour"))
+/// #         );
+/// #     }
+/// # }
+/// ```
+pub fn create_simple_redirect<L: Into<Cow<'static, str>>>(
```
bradleybeddoes: I'm thinking we should have explicit functions here as opposed to giving a bool so much 'power' that would be very easy to trip over.

`create_temporary_redirect` and `create_permanent_redirect` would be my suggestion for the public API (utilising the same private function internally for that DRY goodness).

gotham/src/http/response/mod.rs
```diff
@@ -69,6 +70,84 @@ pub fn create_response(state: &State, status: StatusCode, body: Option<Body>) ->
...
+    permanent: bool,
+    location: L,
+) -> Response {
+    let status = if permanent {
+        StatusCode::PermanentRedirect
+    } else {
+        StatusCode::TemporaryRedirect
+    };
+    let mut res = Response::new().with_status(status);
+    set_headers(&state, &mut res, None, None);
```
bradleybeddoes: Most of the headers set by `set_headers` are not relevant for a 301|302 response.

I'd suggest setting the `XRequestId` header (perhaps break this out into a small function used here and in `set_headers`) only.

gotham/src/http/response/mod.rs
```diff
@@ -69,6 +70,102 @@ pub fn create_response(state: &State, status: StatusCode, body: Option<Body>) ->
...
+    state: &State,
+    location: L,
+) -> Response {
+    let mut res = Response::new().with_status(StatusCode::PermanentRedirect);
+    {
+        let headers = res.headers_mut();
+        headers.set(XRequestId(request_id(state).into()));
+        headers.set(ContentLength(0));
+        headers.set(Location::new(location));
+    }
```
bradleybeddoes: Propose this gets extracted into `fn set_redirect_headers` which can be used here and at L160 to DRY implementation.

gotham/src/http/response/mod.rs
```diff
@@ -69,6 +70,102 @@ pub fn create_response(state: &State, status: StatusCode, body: Option<Body>) ->
...
+/// ```
+pub fn create_permanent_redirect<L: Into<Cow<'static, str>>>(
+    state: &State,
+    location: L,
+) -> Response {
+    let mut res = Response::new().with_status(StatusCode::PermanentRedirect);
+    {
+        let headers = res.headers_mut();
+        headers.set(XRequestId(request_id(state).into()));
+        headers.set(ContentLength(0));
```
bradleybeddoes: AFAIK `ContentLength` isn't valid/required where this is no body content, this could be safely removed.

gotham/src/http/response/mod.rs
onelson: Not sure where my comments on this went, but that sounds good. I wasn't sure about the value of extracting a helper just for this, but if it's your wish I'm happy to oblige.

onelson: extracting the helper feels good. It'll set up folks that want to do something more exotic (like sending bodies with their redirect responses, with one of the other 3xx statuses).

bradleybeddoes:
Awesome, let me just give Travis ye ol "kick in the nuts" and we should be good for merge.

onelson:
Thanks for carrying me through this!

Looked at travis, and I can't explain it. Looks like OOM-killer snapping its jaws down on the job. Surely unrelated, right?

bradleybeddoes:
Not caused by your code @onelson, we're struggling with Travis a lot, seems to be hitting these kinds of strange test run errors quite often but locally we never have an issue. 

Still trying to figure it out.

bradleybeddoes:
This has passed for stable on travis-ci. Change itself is fine, Travis issues still being worked through.

PR #193, opened 2018-03-26
Update to rustfmt-nightly 0.4.1 formatting

smangelsdorf:
As 0.4.1 now ships with nightly rust, we need to update to avoid CI failures.

The changes here shouldn't cause any grief with 0.3.8, since it's just reordering `use` / `mod` statements.

Hopefully once Rust 1.25 ships we can settle on a more stable version of `rustfmt`.

PR #194, opened 2018-03-27
Rewrite `From<MethodSet> for Vec<Method>` impl

smangelsdorf:
The existing implementation, which relied heavily on chained iterators, was causing the optimiser to take a very long time when building in release mode.


bradleybeddoes:
Failures on Travis CI are for rustfmt only which is being considered in #193 currently and waiting for Rust 1.25 to assist decision making.

bradleybeddoes:
Awesome diagnosis work @smangelsdorf 

impressive

PR #195, opened 2018-03-30
Add content type route matcher

ParisLiakos:
Useful for REST APIs that accept a POST or PUT request with content type requirements.

### DIFF
--- a/gotham/src/router/route/matcher/content_type.rs
+++ b/gotham/src/router/route/matcher/content_type.rs
@@ -9,8 +9,8 @@ use router::route::RouteMatcher;
 use state::{request_id, FromState, State};
 
 /// A `RouteMatcher` that succeeds when the `Request` has been made with a `Content-Type` header
-/// that includes one or more supported media types. If the Content-Type is missing
-/// the matcher will fail.
+/// that includes a supported media type. The matcher will fail if the Content-Type
+/// header is missing.
 ///
 /// # Examples
 ///
@@ -39,7 +39,7 @@ use state::{request_id, FromState, State};
 /// state.put(headers);
 /// assert!(matcher.is_match(&state).is_ok());
 ///
-/// // Not a valid Conent-Type header
+/// // Not a valid Content-Type header
 /// let mut headers = Headers::new();
 /// headers.set(ContentType::text());
 /// state.put(headers);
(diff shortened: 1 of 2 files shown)


gotham/src/router/route/matcher/content_type.rs
```diff
@@ -0,0 +1,92 @@
...
+            supported_media_types,
+        }
+    }
+}
+
+impl RouteMatcher for ContentTypeHeaderRouteMatcher {
+    /// Determines if the `Request` was made using a `Content-Type` header that includes one or more
+    /// supported media types. A missing `Content-Type` header, or the value of `*/*` will not match.
+    fn is_match(&self, state: &State) -> Result<(), RouteNonMatch> {
+        // Request method is valid, ensure valid Accept header
```
bradleybeddoes: Just a minor invalid comment (which I note needs fixing in AcceptHeaderRouteMatcher too)

gotham/src/router/route/matcher/content_type.rs
```diff
@@ -0,0 +1,92 @@
...
+        let headers = Headers::borrow_from(state);
+        match headers.get::<ContentType>() {
+            Some(content_type) => {
+                if self.supported_media_types.contains(&content_type.0) {
```
bradleybeddoes: Forgive me, I have a head cold so my brain is not working well, but I'm just not seeing how the `.0` accessor is being used successfully here? (obviously it is!)

gotham/src/router/route/matcher/content_type.rs
```diff
@@ -0,0 +1,92 @@
...
+use hyper::StatusCode;
+use hyper::header::{ContentType, Headers};
+use mime;
+
+use router::non_match::RouteNonMatch;
+use router::route::RouteMatcher;
+use state::{request_id, FromState, State};
+
+/// A `RouteMatcher` that succeeds when the `Request` has been made with a `Content-Type` header
+/// that includes one or more supported media types. If the Content-Type is missing
```
bradleybeddoes: worth noting that `*/*` fails as well (as you have later on)

gotham/src/router/route/matcher/content_type.rs
```diff
@@ -0,0 +1,92 @@
...
+/// #
+/// #   State::with_new(|state| {
+/// #
+/// let supported_media_types = vec![mime::APPLICATION_JSON];
+/// let matcher = ContentTypeHeaderRouteMatcher::new(supported_media_types);
+///
+/// // No content type header
+/// state.put(Headers::new());
+/// assert!(matcher.is_match(&state).is_err());
+///
```
bradleybeddoes: Suggest adding example showing `*/*` being in error also

gotham/src/router/route/matcher/content_type.rs
ParisLiakos: If you see Content type's signature, it is just wrapping a Mime
`pub struct ContentType(pub Mime);`
Its `Display` implementation also uses the same way to access the Mime, see
https://docs.rs/hyper/0.11.24/src/hyper/header/common/mod.rs.html#306

ParisLiakos:
Thanks for the review! i implemented your suggestion but then realized that wildcard `Content-Type` makes no sense. I googled around and also found this; 
https://www.w3.org/Protocols/HTTP/HTRQ_Headers.html

Probably having wildcards in it is invalid.
So, i removed references to wildcards and further adjusted docs

bradleybeddoes:
Failures due to travis timeouts, not issues with this code per se, approving.

PR #198, opened 2018-04-01
Add example of a stateful handler

illicitonion:
And some guidance for when not to use one.

### DIFF
--- a/examples/handlers/stateful/src/main.rs
+++ b/examples/handlers/stateful/src/main.rs
@@ -76,10 +76,7 @@ impl NewHandler for CountingHandler {
 
 /// Create a `Router`
 fn router() -> Router {
-    build_simple_router(|route| {
-        // Each time a request is made to /, call handler.get_handler
-        route.get("/").to_new_handler(CountingHandler::new())
-    })
+    build_simple_router(|route| route.get("/").to_new_handler(CountingHandler::new()))
 }
 
 /// Start a server and use a `Router` to dispatch requests
@@ -95,7 +92,7 @@ mod tests {
     use gotham::test::TestServer;
 
     #[test]
-    fn cookie_is_set_and_counter_increments() {
+    fn counter_increments_per_request() {
         let test_server = TestServer::new(router()).unwrap();
         let response = test_server
             .client()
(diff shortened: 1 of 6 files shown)


bradleybeddoes:
Any chance you could rebase this off master? It'll pick up fixes for Travis that I've put in place so we can be more confident of the change here.

illicitonion:
Done :)

examples/handlers/stateful/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl CountingHandler {
+    fn new() -> CountingHandler {
+        CountingHandler {
+            started_at: SystemTime::now(),
+            visits: Arc::new(Mutex::new(0)),
+        }
+    }
+
+    fn get_handler(&self, state: State) -> (State, Response) {
```
bradleybeddoes: Sorry for the delay on getting to look further at this PR.

I'd suggest implementing [Handler](https://docs.rs/gotham/0.2.1/gotham/handler/trait.Handler.html) for `CountingHandler` to do it "the Gotham way".

You'd then be able to simplify the router definition to:

`route.get("/").to(CountingHandler)`

(On the off chance you need it you could define [NewHandler](https://docs.rs/gotham/0.2.1/gotham/handler/trait.NewHandler.html) and then make use of `to_new_handler`).

illicitonion: Thanks for taking a look!

Unfortunately I can't use `.to` because `CountingHandler` can't be `Clone`, because it contains non-Copyable state (the `Arc<Mutex<usize>>`).

I've updated the code to use `Handler` and `NewHandler`, but it makes me slightly sad because it forces an additional `clone()` of the handler per request (because `Handler.handle` takes `self` not `&self`) - not the end of the world, but something I feel should be avoidable. Is there a gotham-idiomatic way to both cheaply store state _and_ avoid having to copy/clone a `Handler` per request?

bradleybeddoes: Agreed that it isn't an optimal situation right now but necessitated.

@smangelsdorf and I will revisit this and see what is possible with tokio/futures/lifetimes as we adapt to the new direction Tokio is taking, likely as part of Hyper 0.12 work.

bradleybeddoes:
@illicitonion just drawing your attention to https://github.com/gotham-rs/gotham/pull/198#discussion_r182256578 as I missed the question initially due to GH collapsing.

Issue #55, opened 2017-09-30
Async static file serving


tforgione:
On your [blog](https://gotham.rs/blog), you say that some important features are on the roadmap, including *async static file serving*. Is there any work on that for the moment ?

bradleybeddoes:
There hasn't been any work put towards this yet. The core team is busily working on 0.2 and external projects and static files aren't something we had slated for 0.2 at this stage.

That said there wasn't yet a tracking issue for static files so lets use this issue as exactly that.

richardanaya:
I'd like to take this on over the break. I have a project that needs this functionality and directory support :) minus well contribute.

bradleybeddoes:
Welcome to the Gotham project @richardanaya, awesome to hear you're keen to take this issue on. In addition to comments on this issue, please feel free to stop by our [Gitter room](https://gitter.im/gotham-rs/gotham) if you need help with the codebase.

One of the aspects of Gotham that we're very much focused on is security, we want to make it hard for web application developers to create something insecure by always defaulting to "the right thing". 

A significant trap that I've seen other frameworks (Rust and non Rust) fall into when dealing with static files is exposing path traversal attack vectors. [OWASP](https://www.owasp.org/index.php/Main_Page) has some [good documentation on what this is](https://www.owasp.org/index.php/Path_Traversal) if you've not come across it before. Not getting this right scares the crap out of me so it will need some focus.

On the Rust side the [module std::path](https://doc.rust-lang.org/std/path/index.html) will be useful here but it alone will not protect against path traversal attacks.

Good luck with the code, I look forward to using this!

bradleybeddoes:
Just thought I'd check in with you @richardanaya, see how you're going on this one.

scottlamb:
I have a couple crates that may be of interest in [this repository](https://github.com/scottlamb/http-entity):

[...]

PR #199, opened 2018-04-04
Static files WIP

colinbankier:
WIP for adding static file support for #55


bradleybeddoes:
Any chance you could rebase this off master? It'll pick up fixes for Travis that I've put in place so we can be more confident of the change here.

colinbankier:
Sure will do.

gotham/src/handler/static_file.rs
```diff
@@ -0,0 +1,246 @@
...
+        })
+        .map(|contents| {
+            let mime_type = mime_for_path(&path);
+            create_response(state, hyper::StatusCode::Ok, Some((contents, mime_type)))
+        })
+        .unwrap_or_else(|err| error_response(state, err))
+}
+
+fn mime_for_path(path: &Path) -> Mime {
+    guess_mime_type_opt(path).unwrap_or_else(|| mime::TEXT_PLAIN)
```
bradleybeddoes: Not sure if `application/octet-stream` might be better default? Thinking being that with `text/plain` most browsers will probably try and render.

gotham/src/handler/static_file.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+impl Handler for FileHandler {
+    fn handle(self, state: State) -> Box<HandlerFuture> {
+        let response = create_file_response(self.path, &state);
+        Box::new(future::ok((state, response)))
+    }
+}
+
+fn create_file_response(path: PathBuf, state: &State) -> hyper::Response {
```
ashley-rose: It might be cool to move this to a pub fn in response, so users could call it with their custom logic.

gotham/src/router/builder/single.rs
```diff
@@ -166,6 +167,97 @@ pub trait DefineSingleRoute {
...
+    /// # fn main() {
+    /// #   let test_server = TestServer::new(router()).unwrap();
+    /// #   let response = test_server.client()
+    /// #       .get("https://example.com/doc.html")
+    /// #       .perform()
+    /// #       .unwrap();
+    /// #   assert_eq!(response.status(), StatusCode::Ok);
+    /// # }
+    /// ```
+    fn to_filesystem(self, root: &str)
```
vitiral: I think `root` should be `AsRef<Path>` instead, right?

vitiral:
LGTM except the minor API point. Thanks for doing this! I'm amazed how _clean_ this code looks :)

gotham/src/handler/static_file.rs
smangelsdorf: Great suggestion, thanks. I've logged this as #206 so we don't hold up this PR discussing the API (which was my first instinct 😅).

gotham/src/router/builder/single.rs
smangelsdorf: If I'm understanding this suggestion correctly, this means we could use all sorts of `Path`, `OsStr`, `String`, `&str`, etc. values here instead of being restricted to `&str` or an automatically coerced `&String`. The practical benefit is that the path could be pulled from elsewhere (e.g. a config struct in a `lazy_static!`) and used directly rather than having to convert it for the `Router`API.

Is that accurate?

vitiral: `AsRef<Path>` has the benefit of allowing you to use anything that can be coerced into `Path`. So yes, that includes `String`, `&str`, `Path`, `PathBuf`, etc.

All types in the rust stdlib that deal with paths use it, for example [`std::fs::create_dir`](https://doc.rust-lang.org/std/fs/fn.create_dir.html)

smangelsdorf: Thanks for confirming. Good idea.

colinbankier: Great suggestion thanks - I'll take a crack at that change. I learn something new everyday :smile: .

illicitonion:
Looks great! Thanks for putting this together! :)

bradleybeddoes:
@colinbankier just a minor `rustfmt` error in CI to address.

Assuming @smangelsdorf doesn't have any outstanding comments I think we're otherwise good to merge this PR.

PR #207, opened 2018-04-12
0.2.1 port to master

smangelsdorf:
This brings #185 and #194 into `master`.


smangelsdorf:
Failure in beta is caused by skeptic / borrow_bag / travis cache. Everything else is green.

Issue #165, opened 2018-02-28
Track Tokio reform and consider what refactoring is required for Gotham to work with Tokio 0.2


smangelsdorf:
This **may** involve changing to `tokio-threadpool`, which would likely require that all our types are `Send` (so that incomplete futures can be passed across threads).

https://github.com/tokio-rs/tokio/pull/141

ChristophWurst:
Here's a list of breaking changes I have found in Hyper [0.12.x](https://github.com/hyperium/hyper/commits/0.12.x) so far:

- [ ] All uses of `Handle` now need to be new-tokio `Handle` (https://github.com/hyperium/hyper/commit/603c3e4d2c6ebb3d495a85c4d9a8b36583eff3fc)
- [ ] `Method`, `Request`, `Response`, `StatusCode`,
  `Version`, and `Uri` have been replaced with types from the `http`
  crate. (https://github.com/hyperium/hyper/commit/f0cdf82fd46d2a166bf36ca83cb81a297ed9a17e)
- [ ] The `hyper::header` module is gone for now. (https://github.com/hyperium/hyper/commit/f0cdf82fd46d2a166bf36ca83cb81a297ed9a17e, https://github.com/hyperium/hyper/issues/1463#issuecomment-381866090)
- [ ] Removed `Client::get`, since it needed to construct a `Request<B>`
  with an empty body. Just use `Client::request` instead. (https://github.com/hyperium/hyper/commit/f0cdf82fd46d2a166bf36ca83cb81a297ed9a17e)

I gave it a shot and updated Gotham `master` to hyper `0.12.x`. While adapting to Tokio changes seemed straight forward, the removed `hyper::header` module caused most errors. I've reported this at [the hyper repo](https://github.com/hyperium/hyper/issues/1463#issuecomment-381866090) because I couldn't find any drop-in replacement for the module, the `http` crate's header types work differently than the old hyper ones.

> which would likely require that all our types are Send

I'm also pretty sure this will be necessary. https://github.com/hyperium/hyper/commit/603c3e4d2c6ebb3d495a85c4d9a8b36583eff3fc did it too.

There was also an obvious naming conflict when adding the `http` crate because gotham has a `http` module too. Should we import it with a different name?

smangelsdorf:
> There was also an obvious naming conflict when adding the `http` crate because gotham has a `http` module too. Should we import it with a different name?

[...]

PR #209, opened 2018-04-14
GothamListener

nyarly:
This is an attempt to make some of the gotham::start logic more generic,
so that it could be reused in a number of contexts.

I've got something like the unix version of that sketched out, *but*
it turns on using `handle.spawn`, which requires that its `Future` be `'static`
(and `Send`), and the GothamListener::Stream type really isn't.

If I understand the problem correctly, I'm not sure how to resolve it.
Furthermore, I think this is a harbinger of things to come: code like this works
now because `core.run` is more forgiving - in fact `tokio_core::reactor::Core`
specifically calls out that its argument needn't be `'static`.

But `tokio::run()` (and all the ways I can see to put a `Future` in the reactor
in tokio-0.1.5) has the same trait bounds and `Handle::spawn`: gotta be `'static`.


gotham/src/.rustfmt.toml
```diff
@@ -0,0 +1 @@
+max_width = 102
```
nyarly: Without this, rustfmt emits errors because it doesn't know how to fold one of the trait bounds, but without folding, it can't be made less that 100 characters. Manual folding gets put back like it was...

gotham/src/lib.rs
```diff
@@ -69,17 +88,73 @@ where
...
+        thread::spawn(move || run_and_serve(listener, protocol, new_handler));
+    }
+
+    run_and_serve(listener, protocol, new_handler);
+}
+
+fn run_and_serve<'a, G, NH>(listener: G, protocol: Arc<Http>, new_handler: Arc<NH>)
+where
+    G: GothamListener,
+    <G as GothamListener>::Stream: futures::Stream<Item = (TcpStream, SocketAddr), Error = io::Error>
```
nyarly: For reference, this is the trait bound.

gotham/src/.rustfmt.toml
nyarly: c.f. https://github.com/rust-lang-nursery/rustfmt/issues/1080

nyarly:
@smangelsdorf It seems like codecov doesn't like this, but I assume that's because it's at a level that would require integration testing to confirm. If there aren't adequate tests, pointers towards what would be would be appreciated.

smangelsdorf:
Understandable @nyarly … This is a notoriously hard piece of code to test, because it sits outside of all the test harnesses I can imagine writing. I'll get to reviewing this PR soon, thanks for all your work on it.

ChristophWurst:
FYI: tokio_core::reactor::Core was removed with the [tokio reform](https://github.com/gotham-rs/gotham/issues/165). Moreover, all the OS-specific startup code is obsoleted by the Tokio runtime as futures are spawned on a thread pool anyway. Ref https://github.com/gotham-rs/gotham/issues/165#issuecomment-382672829

nyarly:
IIUC, the tokio reform hasn't landed in stable Rust yet? And while I know that will happen, and there's strong motivation for it to happen soon, we don't know *when* that will be.

I'm +1 on dropping all of this once we can get to `tokio` and `hyper-0.12`, but until that day...

gotham/src/os/windows.rs
```diff
@@ -1,152 +1,108 @@
...
+/// Constructs a GothamListener to handle incoming TCP connections.
+///
+/// Note: On Windows this function spawns an extra thread to handle
+/// accepting connections.
+pub fn new_gotham_listener(addr: SocketAddr) -> SocketQueue {
+    let tcp = TcpListener::bind(addr).expect("unable to open TCP listener");
+    let queue = SocketQueue::new(tcp, addr);
+    {
+        let queue = queue.clone();
+        thread::spawn(move || queue.listen());
```
smangelsdorf: Is it fair to say that `tcp` and `addr` could be arguments to this `queue.listen()` call rather than being stored in the `SocketQueue` struct? Having an `Arc<TcpListener>` there seems to give the wrong impression at a casual read – that the listener is thread-safe – when it is not. Refactoring this might help prevent confusion. Thoughts?

gotham/src/lib.rs
```diff
@@ -50,11 +50,30 @@ pub mod state;
...
+
+use service::GothamService;
 use handler::NewHandler;
 
+pub use os::current::new_gotham_listener;
+
+/// Abstracts over TCPListener to provide OS independence for handling incoming TCP connections.
+pub trait GothamListener {
+    /// The type for incoming stream of TCP connections.
+    type Stream;
```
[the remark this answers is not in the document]
nyarly: Oh! You know, I hadn't taken that meaning from the Rust book. I should submit a PR there.

gotham/src/os/unix.rs
```diff
@@ -1,74 +1,34 @@
...
-        let listener = listener.try_clone().expect("unable to clone TCP listener");
-        let protocol = protocol.clone();
-        let new_handler = new_handler.clone();
-        thread::spawn(move || start_core(listener, &addr, &protocol, new_handler));
-    }
-
-    start_core(listener, &addr, &protocol, new_handler);
+#[derive(Clone)]
+pub struct Listener {
+    wrapped: Arc<TcpListener>,
```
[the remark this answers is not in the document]
nyarly: TcpListener has `try_clone` but not `clone`. I've implemented a (possibly panicking) `Clone` on `Listener` using that. The `try_clone().unwrap()` has to happen anyway in order to put the `std::net::TcpListener` in a `tokio_core::net::TcpListener` so it's probably okay?

gotham/src/os/windows.rs
smangelsdorf: > The `SocketQueue` winds up collecting several values, and my first instinct is to gather them together in the constructor.

I can appreciate that instinct. I was coming at it from the other direction, where I know they're only used in a single place and by a single thread (out of necessity, because the listener can't be shared).

I note that your current revision doesn't build due to an issue with the import. I thought that an `Rc` might cause problems here by making the `SocketQueue` struct not `Send + Sync`, though I could be wrong about that.

If the `Rc` version works, it does the "right thing" by preventing the listener being shared.

nyarly:
What do you think of that variation? I think the rest of `SocketQueue` is threadsafe?

I'm a little confused that `std::net::TcpListener` on windows is `Send` if it's not threadsafe. Is that a known issue? (At least, I assume that's the case, since the doc for `Arc` says that it's `Sync + Send` where its contents are...)

smangelsdorf:
Looks good, thanks. I'll do a little more testing but if I don't hit any problems I'll merge it.

> I'm a little confused that `std::net::TcpListener` on windows is `Send` if it's not threadsafe. Is that a known issue?

(Nitpick: The problem is that it's `Sync`)

I'll admit I don't fully understand the source of the problem here. Although I've spent time looking into it and written something that works, I couldn't find any definitive reason for why the threaded version didn't work.

I have two theories I haven't been able to confirm:

* If two threads attempt to accept a connection simultaneously, the kernel won't allow it and will instead return an error; or
* If two threads attempt to *asynchronously* accept a connection, the kernel can't correctly handle the fact that an event is desired by two threads, so it instead gives errors.

The fact that `TcpListener` is `Sync` is a grey area in my opinion. There *may* be usages of the API that are properly thread safe. The API is low-level enough that setting the right socket options might make it work in a way that fits the `Sync` trait.

In cases where you're using it "incorrectly" (as defined by the kernel), the socket does not have any internal inconsistency and instead just returns the errors from the kernel in the `std::io::Result<_>` return types. In that way it can also be seen as thread-safe (the API can be "safely" called across many threads, and won't get anything into an invalid state).

gotham/src/lib.rs
```diff
@@ -69,17 +88,69 @@ where
...
+    NH: NewHandler + 'static,
+{
+    let mut core = Core::new().expect("unable to spawn tokio reactor");
+    let handle = core.handle();
+
+    serve(listener, protocol, new_handler, handle);
+
+    core.run::<future::FutureResult<(), ()>>(future::ok(()))
+        .expect("unable to run reactor over listener");
+}
```
smangelsdorf: This is returning immediately for me, so an application exits immediately without serving any requests. I suspect what's going on here is that the event loop only runs until the `future::ok(())` completes (immediately) and then terminates.

I think `pub fn serve` will need to return a `Future<Item = ()>` to be passed to `core.run()` so that it continues to process connections.

e.g. The way it works in the current `master` branch:
https://github.com/gotham-rs/gotham/blob/f8d03d6e879a071b7e7d341de41af0b4fe0a3710/gotham/src/os/unix.rs#L48-49

nyarly:
I'm closing this for the time being. I'd like to bring it back once we have Tokio and Hyper upgraded.

Issue #208, opened 2018-04-13
Remove skeptic from borrow_bag


bradleybeddoes:
This seems to be root cause of our TravisCI failures currently, we can achieve what it provides with module level doc tests.

PR #211, opened 2018-04-16
Remove unnecessary futures from hello-world

whitfin:
Unless I'm mistaken (which is possible, because I'm new to the project), the `futures` crate is unnecessary to run many of the examples. This PR removes it so that beginners don't end up down the futures rabbit hole too quickly (I myself was confused at the inclusion for a while).

If this is accepted & merged, I'll file the same for the `gotham.rs` repository.


whitfin:
@ChristophWurst cool, once merged I'll file the same PR for the website documentation example. 

Travis appears to be failing the beta build, but it's probably entirely unrelated to this? Just want to check.

ChristophWurst:
> Edit: there are a couple of other examples with the crate too... going to remove them too.

FYI: this is @bradleybeddoes and @smangelsdorf's call. I'm not a maintainer of this crate, so let's wait for their feedback :wink:

whitfin:
@ChristophWurst fair enough, thanks for your feedback :)

I updated this PR to cover all cases of `futures` being unnecessarily included in examples (and compiled them all again, just to check). I also filed a PR for the website too, in case this is merged, to keep them in sync (https://github.com/gotham-rs/gotham.rs/pull/9).

bradleybeddoes:
Thanks for this @whitfin, always happy to remove unintended complexity for newcomers as you've done here.

Beta is certainly not the fault of this PR, we need to implement #208 to sort that out, I'm happy to merge this in the interim however.

PR #212, opened 2018-04-17
Add route matcher for associations

### LINKED ISSUE #196: Make possible to add_route_matcher() to associations
Over in #181 `add_route_matcher()` was made possible for single routes, which is really awesome, improves the DX a lot.
I would like to suggest making it possible for associations too.
It would be really awesome if we could do something like this

```rust
route.associate("/api/v1/users", |assoc| {
   let json_matcher = AcceptHeaderRouteMatcher::new(vec![mime::APPLICATION_JSON]);
   assoc.add_route_matcher(json_request_matcher);

   assoc.get()
      .to(users::get_users);

    assoc.post()
      .to(users::create_user);

    // Maybe put and delete implementations.
});
```
So that we dont have to specify the matcher over and over again for all single routes of the association.

I havent actually dig into the code to see how easy that would be..Thinking out loud mostly.

bradleybeddoes:
Permits `RouteMatcher` instances to be defined for, and then applied to, associated routes.

Closes #196


bradleybeddoes:
CI fail appears to be skeptic again, which I've addressed now in #213.

Issue #71, opened 2017-12-16
Simplify use statements with re-exports


IntrepidPig:
Right now, gotham is structured with a relatively deep tree. This leads to lots of large `use` statements in certain files, for example, the [routing file in the example app](https://github.com/gotham-rs/example-app/blob/master/src/boot/router.rs). These statements can be simplified by reexporting types deeper in the tree from higher level modules.

In case I'm not explaining it properly, I mean being able to convert this:
```rust
use gotham::router::request::path::NoopPathExtractor;
use gotham::router::request::query_string::NoopQueryStringExtractor;
use gotham::router::tree::TreeBuilder;
use gotham::router::tree::node::{NodeBuilder, SegmentType};
```
to something like
```rust
use gotham::router::request::{NoopPathExtractor, NoopQueryStringExtractor};
use gotham::router::tree::{TreeBuilder, NodeBuilder, SegmentType};
```

I think this could also be done with other crates like hyper and mime to provide a more streamlined experience, but that might be too much.

bradleybeddoes:
I'm certainly keen to see this happen, I guess the thought was to leave this until after 0.1 (possibly 0.2?) so that folks aren't impacted by structure that could be in flux as they upgrade releases.

However as we've seen in https://github.com/gotham-rs/gotham/pull/70 folks seem comfortable with this kind of thing.

@smangelsdorf I'm happy to progress this if you're onboard.

smangelsdorf:
Sounds good.

smangelsdorf:
Reopening. That PR wasn't a complete fix.

PR #214, opened 2018-04-19
Rename gotham::http to gotham::protocol

ChristophWurst:
This prevents naming conflicts of the `http` crate used in Hyper 0.12.

As [suggested](https://github.com/gotham-rs/gotham/issues/165#issuecomment-382529429) by @smangelsdorf, I chose `protocol` as the new module name.

There were 6 doc tests that failed locally. Errors like this one popped up, although I fixed `gotham_derive`:

```
---- src/router/builder/single.rs - router::builder::single::DefineSingleRoute::with_query_string_extractor (line 243) stdout ----
        error[E0433]: failed to resolve. Could not find `http` in `gotham`
  --> src/router/builder/single.rs:261:34
   |
19 | #[derive(StateData, Deserialize, StaticResponseExtender)]
   |                                  ^^^^^^^^^^^^^^^^^^^^^^ Could not find `http` in `gotham`
```


ChristophWurst:
> There were 6 doc tests that failed locally.

They fail on Travis as well. I still don't know where the old `http` module is referenced at.
```
failures:
    src/extractor/path.rs - extractor::path::PathExtractor (line 21)
    src/extractor/query_string.rs - extractor::query_string::QueryStringExtractor (line 21)
    src/router/builder/mod.rs - router::builder::AssociatedRouteBuilder<'a, C, P, PE, QSE>::with_path_extractor (line 348)
    src/router/builder/mod.rs - router::builder::AssociatedRouteBuilder<'a, C, P, PE, QSE>::with_query_string_extractor (line 409)
    src/router/builder/single.rs - router::builder::single::DefineSingleRoute::with_path_extractor (line 174)
    src/router/builder/single.rs - router::builder::single::DefineSingleRoute::with_query_string_extractor (line 243)
```

whitfin:
@ChristophWurst small note, I added another example recently in https://github.com/gotham-rs/gotham/pull/215. Can you please grab the `gotham::http` in that example too?

bradleybeddoes:
> Incidentally, your issue with those tests is because the repository gotham points at the published 0.2 of gotham_derive, not the version in the repo (for whatever reason).

Good catch, it shouldn't be, leftover from our migration to a single workspace. [This is a bug](https://github.com/gotham-rs/gotham/blob/master/gotham/Cargo.toml#L36) that can get sorted as part of this PR if you wouldn't mind @ChristophWurst.

ChristophWurst:
Fixed :rocket:

bradleybeddoes:
@smangelsdorf and I have been talking about naming offline (I was late to this entire discussion).

I felt like `gotham::protocol::{request, response, header}` wasn't the best naming, as in which protocol are we referring to? This confusion would be compounded when Gotham does something with WebSockets.

I believe we arrived at changing `gotham::http`/`gotham::protocol` to `gotham::helpers::http` which was inspired by the existing [module comment](https://github.com/gotham-rs/gotham/blob/master/gotham/src/http/mod.rs#L1) and to my mind properly describes what the code here does. As part of this `{request, response, header}` components would be re-exported under `gotham::helpers::http` keeping #71 in mind.

Thoughts?

smangelsdorf:
> I believe we arrived at changing `gotham::http`/`gotham::protocol` to `gotham::helpers::http`

Yep, that's where we got to.

PR #215, opened 2018-04-19
Add an example of matching Regex patterns in routes

whitfin:
Based on a conversation in Gitter, there isn't any documentation around Regex patterns that I can see, and it's not particularly obvious. This will add an example to cover at least the basic use cases.


bradleybeddoes:
Thanks very much!!

PR #216, opened 2018-04-25
[WIP] Tokio reform

ChristophWurst:
To do
- [ ] Update to hyper 0.12 https://github.com/gotham-rs/gotham/issues/165#issuecomment-381868319
- [ ] Fix examples

If you comment out the `test` module in `lib.rs` you can already compile and test the hello world example.


nyarly:
@ChristophWurst What's the status here? My understanding is that we can probably hold 0.12 until later (and that they're waiting for something to land in stable... I'm unclear of the details) but that `tokio` should be ready to go, and that group has pretty well committed to the switch. From what it looks like, migrating to tokio would be a huge boon.

nyarly:
Okay, I might well do that unless @smangelsdorf or @bradleybeddoes have any objections.

It's my understanding that `hyper 0.11` is compatible with `tokio`, so that conversion doesn't have to happen yet. There are a ton of ergonomics to be had, including `Handle::current()` (name?).

ChristophWurst:
> It's my understanding that hyper 0.11 is compatible with tokio

I'm really not sure about that. If you take a look at https://github.com/hyperium/hyper/issues/1443, the comments indicate that the switch to `tokio` will happen only for 0.12 since it's a breaking change. Moreover, there is [no API to get hold of the old handle](https://docs.rs/tokio/0.1.6/tokio/?search=handle) in `tokio`.

ChristophWurst:
I'm closing this PR as I don't see any interest of the current maintainers to integrate my effort. If any of the [new maintainer](https://gotham.rs/blog/2018/05/31/the-state-of-gotham.html) (if any) want to build on this, feel free to take my commits and finish the last missing pieces of the migration.

ChristophWurst:
> @ChristophWurst just to clarify, at this point is this PR based solely on Tokio 0.1, or has Hyper also changed?

That was kind of my intention, but you need a `Handle` to create a hyper client (used esp. in the `tests` module) which you won't get from the `tokio` lib. I think both have to be migrated at the same time.

nyarly:
I need to review more deeply, but one of the advantages of `tokio` over `-core` that I recall was that it's *easier* to get `Handle`s now. IIRC you can promote `tokio-core` `Handle`s to `tokio` ones, right?

ChristophWurst:
> IIRC you can promote tokio-core Handles to tokio ones, right?

Nope, just in the opposite direction. See https://github.com/gotham-rs/gotham/pull/216#issuecomment-388915573

nyarly:
I'm really confused now: Tokio Core is the old thing, Just Tokio is new. So if Tokio Core has an API to create Tokio Handles, aren't we saying the same thing?

That said, the method you link is in `tokio-reactor` and produces `-reactor` `Handle`s. I guess I know what my reading list is like for the forseeable future.

PR #221, opened 2018-05-07
Add a basic CLF request logger

whitfin:
This PR introduces a basic CLF request logger. It's based on something I wrote for a work project, but I cleaned it up for public use and figured it'd be worth PR'ing seeing as people like request logs. It's nothing particularly special, but it's better than nothing.


whitfin:
For some reason when I point at the branch to pull this in, I see weird issues with the following:

```
error[E0277]: the trait bound `gotham_middleware_logger::LoggingMiddleware: gotham::middleware::NewMiddleware` is not satisfied
  --> src/router.rs:19:10
   |
19 |         .add(LoggingMiddleware::with_duration(Level::Info, true))
   |          ^^^ the trait `gotham::middleware::NewMiddleware` is not implemented for `gotham_middleware_logger::LoggingMiddleware`
```

But doing this inside the repo as a test is just fine... is there some scoping with `NewMiddleware` that I don't know about?

whitfin:
I'm also not sure if I should be using `#[derive(NewMiddleware)]`, when I could do this:

```rust
impl NewMiddleware for LoggingMiddleware {
    type Instance = LoggingMiddleware;

    fn new_middleware(&self) -> io::Result<Self::Instance> {
        Ok(LoggingMiddleware { ..*self })
    }
}
```

Is this block better/faster than the clone? Intuition says yes, but perhaps that's not the case.

smangelsdorf:
> is there some scoping with NewMiddleware that I don't know about?

This will be caused by multiple versions of the Gotham crate existing in your dependency hierarchy.

I'm using Gotham from git `master` in a project (which is a cargo workspace, not just a single crate), and the following dependency configuration works for me:

```toml
# In the gotham app crate's `Cargo.toml`.
[dependencies]
gotham = { git = "https://github.com/gotham-rs/gotham" }
gotham_derive = { git = "https://github.com/gotham-rs/gotham" }
gotham_middleware_diesel = { git = "https://github.com/gotham-rs/gotham" }
```

```toml
# In the top-level workspace `Cargo.toml`
[patch.crates-io]
gotham = { git = "https://github.com/gotham-rs/gotham" }
gotham_derive = { git = "https://github.com/gotham-rs/gotham" }
gotham_middleware_diesel = { git = "https://github.com/gotham-rs/gotham" }

whitfin:
@smangelsdorf that's fine then - I just wanted to make sure I didn't have a bug pushed up in this PR :) In that case, this is ready for reviewing - seems complete enough!

whitfin:
I think I'm going to re-target this for the `middleware` module inside Gotham itself.

whitfin:
Squashed and embedded into the internal middleware rather than as a separate crate!

PR #222, opened 2018-05-08
Add names to threads spawned by Gotham.

im-0:
This helps to distinguish log entries from different threads in the log file.


im-0:
Now my email is right in commit.

PR #231, opened 2018-06-04
[WIP] Migration to Hyper v0.12

whitfin:
This PR includes migration to the new version of Tokio, via the `tokio` crate. It includes an update to Hyper 0.12, seeing as that's required for Hyper to function with Tokio correctly.


gotham/src/os/mod.rs
```diff
@@ -1,9 +0,0 @@
-#[cfg(not(windows))]
```
nyarly: This makes me super happy.

gotham/src/test/mod.rs
```diff
@@ -173,13 +171,19 @@ where
...
 
-        handle.spawn(f);
+        {
+            self.data.runtime.read().unwrap().spawn(f);
+        }
 
-        let client = Client::configure()
-            .connector(TestConnect {
-                stream: cell::RefCell::new(Some(cs)),
+        let client = Core::new()
```
nyarly: This looks like the crucial remaining usage of tokio_core, right?

whitfin: @nyarly yep! This is the one that has to stick around until Hyper 0.12 is adopted.

whitfin:
This is getting there very slowly; FWIW once it actually compiles, we should go over it to clean up before merging - I'm very much taking a quick approach to just get something working for now.

whitfin:
Closing in favour of #246.

PR #234, opened 2018-06-14
Migrate to Tokio for main runtime

whitfin:
This PR will introduce Tokio Reform for the main Gotham Runtime. The main advantage here is that it'll strip out all OS-specific code, in favour of a uniform socket using Tokio.

This does **not** remove `tokio_core` completely, as it's still required in `test/mod.rs` for Hyper v0.11. This will be removed in a follow up when we migrate to Hyper v0.12. For now, the two can live together in peace seeing as (almost?) all usage is internal.


whitfin:
There are two failing examples because of the use of `Handle` that I don't think we can do anything about until we migrate Hyper (because they use Hyper).

whitfin:
@smangelsdorf and @nyarly, 

If we're going to ship Tokio Reform separately from Hyper 0.12, we need to decide what to do about the `async_handlers` example as it makes use of Hyper 0.11 (hence the breaking CI builds).

nyarly:
Ran `cargo fmt` since that was blocking Travis on Stable.

PR #246, opened 2018-07-06
Hyper v0.12 Adoption

### LINKED ISSUE #165: Track Tokio reform and consider what refactoring is required for Gotham to work with Tokio 0.2

nyarly:
This is close to compiling - we'll see how tests go.

I'll come back to it soon, but I wanted to make it more public in case anyone has ideas about what needs to happen to fix it.

This will fix #165.


examples/handlers/async_handlers/Cargo.toml
```diff
@@ -1,3 +1,4 @@
+
```
whitfin: Seems like this was some accidental spacing.

gotham/Cargo.toml
```diff
@@ -32,6 +33,10 @@ linked-hash-map = "0.5"
 num_cpus = "1.8"
 crossbeam = "0.3"
 regex = "1.0"
+cookie = "0.10"
+http = "0.1.7"
```
whitfin: I think it'd be better if we stuck to `x.x` minor for packages, so that:

```
http = "0.1"
failure = "0.1"
failure_derive = "0.1"
```

Just typically easier to pick up bugfixes that way.

gotham/src/helpers/http/mod.rs
```diff
@@ -1,6 +1,6 @@
 //! Helpers for HTTP request handling and response generation
 
-pub mod header;
+//pub mod header;
```
whitfin: This can be dropped because I think they've all been removed!

gotham/src/helpers/http/response/mod.rs
```diff
@@ -1,14 +1,15 @@
...
+                    X_CONTENT_TYPE_OPTIONS, X_FRAME_OPTIONS, X_XSS_PROTECTION};
+use hyper::{Body, Method, Response, StatusCode};
 use mime::Mime;
 use std::borrow::Cow;
 
-use helpers::http::header::{XContentTypeOptions, XFrameOptions, XRequestId, XXssProtection};
 use state::{request_id, FromState, State};
 
-type Body = (Vec<u8>, Mime);
+type DataMime = (Vec<u8>, Mime);
```
whitfin: This is a better name, but I question whether we really need this type. It's only used a couple of times?

gotham/src/helpers/http/response/mod.rs
```diff
@@ -64,10 +64,22 @@ type Body = (Vec<u8>, Mime);
...
+    let mime = body.clone().map(|(_, mime)| mime);
+
+    extend_response(state, status, &mut builder, mime);
+    match body {
+        Some((body, _)) => match *Method::borrow_from(state) {
+            Method::HEAD => builder.body(Body::empty()),
+            _ => builder.body(body.into()),
+        },
+        None => builder.body(Body::empty()),
+    }.unwrap()
```
whitfin: It's unfortunate that we have to stay compatible here, because it would be nice to return a `Result` rather than unwrapping here.

gotham/src/helpers/http/response/mod.rs
```diff
@@ -64,10 +64,22 @@ type Body = (Vec<u8>, Mime);
...
-    let mut res = Response::new();
-    extend_response(state, &mut res, status, body);
-    res
+pub fn create_response(
+    state: &State,
+    status: StatusCode,
+    body: Option<DataMime>,
+) -> Response<Body> {
+    let mut builder = Response::builder();
+    let mime = body.clone().map(|(_, mime)| mime);
```
whitfin: I'm not sure if this can be done in such a way as to avoid the clone; maybe unpack it and clone the mime type itself? The vec in body could be potentially quite large.

gotham/src/helpers/http/response/mod.rs
whitfin: Having said that, if this is `0.3.0`, maybe we can change this?

gotham/src/helpers/http/response/mod.rs
```diff
@@ -140,18 +155,21 @@ pub fn create_permanent_redirect<L: Into<Cow<'static, str>>>(
...
+pub fn create_temporary_redirect<B: Default, L: Into<Cow<'static, str>>>(
     state: &State,
     location: L,
-) -> Response {
-    let mut res = Response::new().with_status(StatusCode::TemporaryRedirect);
+) -> Response<B> {
+    let mut res = Response::builder()
+        .status(StatusCode::TEMPORARY_REDIRECT)
+        .body(B::default())
+        .unwrap();
```
whitfin: Same here with an `unwrap` we might want to avoid.

gotham/src/helpers/http/response/mod.rs
```diff
@@ -225,21 +247,14 @@ pub fn extend_response(state: &State, res: &mut Response, status: StatusCode, bo
...
-        None => {
-            set_headers(state, res, None, None);
-            res.set_status(status);
-        }
-    };
+    match mime {
+        Some(mime) => builder.header(CONTENT_TYPE, mime.as_ref()),
+        None => builder,
+    }.header(
+        HeaderName::from_lowercase(b"x-request-id").unwrap(),
```
whitfin: I thought about this too, I'm not sure if we can do something better for the default headers we want to attach. But maybe we can solve that later!

gotham/src/helpers/http/response/mod.rs
```diff
@@ -225,21 +247,14 @@ pub fn extend_response(state: &State, res: &mut Response, status: StatusCode, bo
...
-                Method::Head => (),
-                _ => res.set_body(body),
-            };
-        }
-        None => {
-            set_headers(state, res, None, None);
-            res.set_status(status);
-        }
-    };
+    match mime {
```
whitfin: This would look a little better like this (IMO, anyway):

```
if let Some(mime) = mime {
  *builder = builder.header(CONTENT_TYPE, mime.as_ref());
}
```

gotham/src/helpers/http/response/mod.rs
```diff
@@ -442,12 +465,20 @@ pub fn set_headers(state: &State, res: &mut Response, mime: Option<Mime>, length
...
 ) {
     let headers = res.headers_mut();
-    headers.set(XRequestId(request_id(state).into()));
-    headers.set(Location::new(location));
+    set_request_id(state, headers);
+    headers.insert(LOCATION, location.into().to_string().parse().unwrap());
+}
+
+/// Sets the request id inside a given `HeaderMap`.
+fn set_request_id(state: &State, headers: &mut HeaderMap) {
```
whitfin: Maybe we can resurrect the `headers` module with default header functions like this for `set/get`. We can use statics for the `HeaderName` then to avoid converting often.

gotham/src/lib.rs
```diff
@@ -40,6 +40,14 @@ extern crate uuid;
 #[macro_use]
 extern crate serde_derive;
 
+extern crate cookie;
+extern crate http;
+
+extern crate failure;
+//#[macro_use]
+//extern crate failure_derive;
```
whitfin: Just a note that this is here in case we want to remove it later.

gotham/src/router/route/matcher/accept.rs
```diff
@@ -91,19 +91,24 @@ impl RouteMatcher for AcceptHeaderRouteMatcher {
...
-        match Headers::borrow_from(state).get::<Accept>() {
+        let headers = HeaderMap::borrow_from(state);
+        match headers.get(ACCEPT) {
             Some(accept) => {
-                for ra in accept.iter().map(|qi| &qi.item) {
-                    if *ra == mime::STAR_STAR || self.supported_media_types.contains(ra) {
-                        return Ok(());
-                    }
+                if accept == "*/*"
+                    || self.supported_media_types.contains(&accept
```
whitfin: I'm not sure but there may be a better way to do this; can't think of one now, just a note for later.

gotham/src/router/route/matcher/content_type.rs
```diff
@@ -72,20 +72,26 @@ impl RouteMatcher for ContentTypeHeaderRouteMatcher {
     /// Determines if the `Request` was made using a `Content-Type` header that includes a
     /// supported media type. A missing `Content-Type` header will not match.
     fn is_match(&self, state: &State) -> Result<(), RouteNonMatch> {
-        match Headers::borrow_from(state).get::<ContentType>() {
+        let headers = HeaderMap::borrow_from(state);
+        match headers.get(CONTENT_TYPE) {
             Some(content_type) => {
-                if self.supported_media_types.contains(&content_type.0) {
+                if self.supported_media_types.contains(&content_type
```
whitfin: Same goes here, we might find a better way to do this.

gotham/src/router/route/mod.rs
```diff
@@ -49,6 +49,7 @@ pub enum Delegation {
 /// `Route` exists as a trait to allow abstraction over the generic types in `RouteImpl`. This
 /// trait should not be implemented outside of Gotham.
 pub trait Route: RefUnwindSafe {
+    type ResBody;
```
whitfin: Nitpicky, but this is probably better as `ResponseBody` just because it's a type. Or maybe just `BodyType` or something. `Body` would be ideal.

I know it's in sync with Hyper, but it might be nice to distance ourselves a little bit here.

gotham/src/service/trap.rs
```diff
@@ -6,14 +6,16 @@ use std::error::Error;
...
 use futures::future::{self, Future, FutureResult, IntoFuture};
 use futures::Async;
-use hyper::{self, Response, StatusCode};
+use hyper::{Body, Response, StatusCode};
 
 use handler::{Handler, HandlerError, IntoResponse, NewHandler};
 use service::timing::Timer;
 use state::{request_id, State};
 
+type AnError = failure::Compat<failure::Error>;
```
whitfin: This name probably needs to be cleaned up to something more specific ;)

gotham/src/state/request_id.rs
```diff
@@ -22,15 +21,14 @@ pub(super) struct RequestId {
 /// that a value for `RequestId` is always available.
 pub(crate) fn set_request_id<'a>(state: &'a mut State) -> &'a str {
     if !state.has::<RequestId>() {
-        let request_id = match Headers::borrow_from(state).get::<XRequestId>() {
+        let request_id = match HeaderMap::borrow_from(state).get("X-Request-ID") {
```
whitfin: Goes back to the comment further up about utilities for headers we "recognise" officially.

gotham/src/test/mod.rs
```diff
@@ -189,45 +168,60 @@ where
...
-            core.run(f.select2(timeout))
-        };
+        let timeout = Delay::new(timeout_duration);
 
-        match run_result {
-            Ok(future::Either::A((item, _))) => Ok(item),
-            Ok(future::Either::B(_)) => Err(TestRequestError::TimedOut),
-            Err(future::Either::A((e, _))) => Err(TestRequestError::HyperError(e)),
-            Err(future::Either::B((e, _))) => Err(TestRequestError::IoError(e)),
+        match self.run_future(f.select2(timeout).map_err(|either| {
```
whitfin: The readability here is a bit... crazy. Might be a nice idea to store the future separately above before the `match`.

gotham/src/test/mod.rs
```diff
@@ -236,118 +230,128 @@ pub struct TestClient<NH>
...
+    connect: Box<Future<Item = TestConnect, Error = CompatError> + Send + Sync>,
     test_server: TestServer<NH>,
 }
 
 impl<NH> TestClient<NH>
 where
     NH: NewHandler + Send + 'static,
+    // + Fn() -> io::Result<H>,
+    //H: Send + FnOnce(state::State) -> IHF,
+    //IHF: IntoHandlerFuture + Sized,
```
whitfin: Just a note that these comments either need to be removed or uncommented.

gotham/src/test/mod.rs
```diff
@@ -435,16 +439,27 @@ impl TestResponse {
...
+
+    fn connect(&self, _dst: Destination) -> Self::Future {
+        Box::new(future::ok((self.stream, Connected::new())))
+    }
 }
 
-impl client::Service for TestConnect {
-    type Request = hyper::Uri;
+/*
+impl service::Service for TestConnect {
```
whitfin: This service can also be removed or uncommented in future.

gotham/src/test/mod.rs
```diff
@@ -546,7 +564,7 @@ mod tests {
             .perform();
 
         match res {
-            Err(TestRequestError::TimedOut) => (),
+            //Err("timed out") => (),
```
whitfin: Another dangling comment.

gotham/Cargo.toml
nyarly: My understanding was that `x.y.z` -> `^x.y.z` -> `>= x.y.z, <x.y+1` For semvered crates, that should cover bugfixes without added features. Of course, for `x=0` all bets are off anyway.

Is my understanding right here? I can definitely see the logic to relaxing to `0.y` even if so, though.

o01eg:
There some typos I sent in PR to PR: https://github.com/nyarly/gotham/pull/1

gotham/src/helpers/http/response/mod.rs
nyarly: I'd be happy to see that tackled in a separate PR. `create_response` gets called in a bunch of places, though.

whitfin:
@nyarly just to bring to your attention that Travis has Clippy warning on a few things to clean up: https://travis-ci.org/gotham-rs/gotham/jobs/414189144

Also the failure with formatting check appears to be caused by a `cargo fmt` issue: https://travis-ci.org/gotham-rs/gotham/jobs/414189143 due to https://github.com/rust-lang/cargo/issues/5208. Not sure how we fix that for now, perhaps if we try remove the `--` from the Travis config it might work? No idea really.

Once the build passes, I'll sweep another review :)

nyarly:
Thanks. I'll follow up there. Now I just need to figure out how to get Clippy installed...

whitfin:
Working on this today and tomorrow, looking good so far. Quick note to @o01eg that your changes have been pulled into this branch.

o01eg:
@whitfin New version of rustfmt accepts other argument: `cargo fmt --all -- --check`

whitfin:
At this point, I've addressed most things and it looks good to go. Build is also passing; there are just a few things I want to run back over quickly before merging. Will update soon!

PR #247, opened 2018-07-06
merge GothamService and GothamConnectedService, closes #241

### LINKED ISSUE #241: Merge GothamService and GothamConnectedService
One of them is unnecessary; it appears that they're separated for the test module - but this can be easily changed and one of them can be removed (probably `GothamConnectedService`).

jxs:


whitfin:
Per @nyarly, the reason we have two services is so that we can use the compiler for guarantees. As such I'm going to close this for now, but thank you for the PR @jxs!

Issue #220, opened 2018-05-07
Is there any way to gracefully terminate the event loop?


im-0:
I need to do some additional cleanup actions after gotham::start(). So it would be very good if there is a way to make it exit. For example, using custom SIGTERM handler (tokio-signal) with some glue code.

ChristophWurst:
A possible solution would be to expose the Hyper `run_until` API in Gotham, too: https://docs.rs/hyper/0.11.26/hyper/server/struct.Server.html#method.run_until

nyarly:
#310 includes @im-0's  example, updated to modern Tokio/Hyper, that demonstrates how to do this.

PR #223, opened 2018-05-10
Implement run_until() similar to hyper's run_until()

im-0:
This adds the ability to gracefully shut down the event loop using special signaling future.

Fixes gotham-rs/gotham#220.

PR #249, opened 2018-07-08
Support graceful shutdown.

im-0:
This is a version of gotham-rs/gotham#223 updated for the current `master` branch.

Fixes gotham-rs/gotham#220.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,6 @@ addons:
 after_success: |
   if [[ "$TRAVIS_RUST_VERSION" == stable && "$SHARD" != rustfmt ]]; then
     bash <(curl https://raw.githubusercontent.com/xd009642/tarpaulin/master/travis-install.sh)
-    cargo tarpaulin --all --out Xml
+    cargo tarpaulin --all --forward --out Xml
     bash <(curl -s https://codecov.io/bash)
   fi
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
 
     ## Examples (these crates are not published)
     "examples/hello_world",
+    "examples/hello_world_until",
 
     # Tera template
     "examples/templating/tera",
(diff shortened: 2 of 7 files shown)


whitfin:
@im-0 I'm not sure I understand exactly what this adds... the ability to CTRL+C without killing the application itself? Can you just brief me real quick? I ask because I saw `run_until` references but they don't seem to be involved here.

im-0:
@whitfin, in general, this adds a way to programmatically stop the running Gotham server. May be used to perform deinitialization when stopping the server (something more complex than just "kill the process").

> I ask because I saw run_until references but they don't seem to be involved here.

Previous version (see gotham-rs/gotham#223) was designed to be similar to the [run_until() from hyper 0.11](https://docs.rs/hyper/0.11.27/hyper/server/struct.Server.html#method.run_until). But as 1) this function was removed from hyper 0.12 and 2) Gotham code was also changed heavily during migration to new `tokio`, I made this new set of patches. New solution is much simpler.

im-0:
(rebased onto current master)

colinbankier:
Thanks - this looks good to me :+1: 
Just thinking out load, I was wondering if the dance to set up the signal etc could be included in `gotham/src/lib.rs` so that it could be re-used rather then each app setting it up - and maybe exposed in a way you could pass a closure in for the shutdown hook. Not really necessary - just a thought.

im-0:
@whitfin will do.

nyarly:
@im-0 Checking in - if I understand your change correctly, the actual Gotham code is superseded by #254, but the example code looks great. Are you still willing to update the example against init_server?

ravernkoh:
Would love this to be merged!

nyarly:
I've submitted #310 to get this example merged. Thank you, @im-0 for you work on this.

@ravernkoh The feature itself is already present in Gotham as it ships today. The example as presented in #310 illustrates how to do that.

ravernkoh:
Thanks @nyarly. Although I feel the issue should only be closed when the PR is merged.

PR #250, opened 2018-07-10
Add Tera template engine usage

### LINKED ISSUE #224: How to return html page, in particular a Tera template, from an action?
I haven't found an example of how to work with html in gotham, is there any?

jxs:
Hi,
If you prefer, I can implement this example using a Middleware instead

This fixes #224.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,9 @@ members = [
     ## Examples (these crates are not published)
     "examples/hello_world",
 
+    # Tera template
+    "examples/templating/tera",
+
     # routing
     "examples/routing/introduction",
     "examples/routing/http_verbs",
--- a/examples/README.md
+++ b/examples/README.md
@@ -40,6 +40,7 @@ information on functionality and ordering.
 | [Handlers](handlers) | Developing application logic that responds to web requests. | 4 |
 | [Middleware](middleware) | Developing custom middleware for your application. | 1 |
 | [Into Response](into_response) | Implementing the Gotham web framework's `IntoResponse` trait. | 1 |
+| [Tera Template](templating/tera) | An example using Tera template engine | 1 |
 
 ^ Gotham web framework examples are under active development.
 
--- a/examples/templating/tera/Cargo.toml
+++ b/examples/templating/tera/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "gotham_examples_tera_templates"
+description = "A Tera template engine usage example application for Gotham."
+version = "0.0.0"
+publish = false
+
+[dependencies]
+gotham = { path = "../../../gotham" }
+
+hyper = "0.11"
+mime = "0.3"
+tera = "0.11.8"
+lazy_static = "1.0"
--- a/examples/templating/tera/README.md
+++ b/examples/templating/tera/README.md
@@ -1,0 +1,63 @@
+# Tera Template
+
+An example usage of Tera template engine working with Gotham.
+
+## Running
+
+From the `examples/tera_template` directory:
+
+```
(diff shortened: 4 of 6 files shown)


im-0:
It seems that you also need to add your new example into `Cargo.toml` at the repository root. Otherwise CI will not pick up and build/test your code.

Also, there is a `examples/README.md` file which contains links for all existing examples with descriptions. Consider adding link for your example there.

(WARNING: I am not the maintainer of this repository and thus I may be wrong =))

jxs:
thanks @im-0 forgot to stage the main cargo.toml, and update the example README.md.

whitfin:
At a first glance this looks fine and is fine as an example :) I'm not sure what a middleware would provide?

jxs:
@whitfin the Middleware would make unnecessary to use a static variable, and instead you would access the tera reference via State, on the other hand it would be more complex

Cargo.toml
```diff
@@ -12,6 +12,9 @@ members = [
     ## Examples (these crates are not published)
     "examples/hello_world",
 
+    #Tera template
```
whitfin: Space after the `#` :)

examples/README.md
```diff
@@ -40,6 +40,7 @@ information on functionality and ordering.
 | [Handlers](handlers) | Developing application logic that responds to web requests. | 4 |
 | [Middleware](middleware) | Developing custom middleware for your application. | 1 |
 | [Into Response](into_response) | Implementing the Gotham web framework's `IntoResponse` trait. | 1 |
+| [Tera Template](tera_template) | An example using Tera template engine | 1 |
```
whitfin: I think this might warrant a new `templating` directory for examples, so that we can add examples for other engines as we need to. Do you want to add this directory (you can base it on one of the other directories). Then your example can just be named `tera`.

examples/tera_template/README.md
```diff
@@ -0,0 +1,63 @@
+# Tera Template
+
+Am example usage of Tera template engine working with Gotham.
```
whitfin: Small typo of "Am". It would also be nice to get a link to the Tera project here too.

examples/tera_template/src/main.rs
```diff
@@ -0,0 +1,74 @@
+//! A Hello World example application for working with Gotham.
```
whitfin: These docs need updating to reflect that they're using Tera.

examples/tera_template/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+#[macro_use]
+extern crate tera;
+
+use hyper::{Response, StatusCode};
+
+use gotham::helpers::http::response::create_response;
+use gotham::state::State;
+use tera::{Context, Tera};
+
+///Assuming the Rust file is at the same level as the templates folder
```
whitfin: Nitpick, but a space after the `///` and don't start the second line with a leading `,`.

examples/tera_template/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+///, we can get a Tera instance that way:
+lazy_static! {
+    pub static ref TERA: Tera = {
+        compile_templates!(concat!(env!("CARGO_MANIFEST_DIR"), "/templates/**/*"))
+    };
+}
+
+/// Create a `Handler` which is invoked when responding to a `Request`.
+///
+/// How does a function become a `Handler`?.
```
whitfin: Maybe replace these last few lines with information about how you're using Tera, and what it's rendering.

jxs:
thanks for the review @whitfin, think everything is alright now , but confirm it please

PR #252, opened 2018-07-15
Add StateMiddleware to simplify shared state

### LINKED ISSUE #251: Provide an example for attaching items to the request state
It's come up often as to how to share state between handlers. There should be an example showing how to do this using a `Middleware`.

whitfin:
This PR will add a new `StateMiddleware` which enables attaching generic types to requests easily enough. It masks the `Middleware` requirement needed for state, which is much more familiar to new users. An example is also included!

This will fix #251.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
 
     ## Examples (these crates are not published)
     "examples/hello_world",
+    "examples/shared_state",
 
     # routing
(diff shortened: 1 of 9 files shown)


colinbankier:
This looks super useful :+1:

whitfin:
@colinbankier do you want to approve/merge this if you're ready for it to go into master?

PR #254, opened 2018-07-15
Enable Runtime customization on startup

### LINKED ISSUE #95: Question: start gotham server using a provided Handle

whitfin:
I see no reason to limit the thread count of the backing `Runtime`, so this will allow you to customize either the thread count, or provide your own runtime entirely. 

In addition to the two extra startup functions, I made a server setup function public to enable developers to spawn it in a totally different way if they wish.

This will fix #95.

### DIFF
--- a/gotham/src/lib.rs
+++ b/gotham/src/lib.rs
@@ -56,7 +56,7 @@ use std::sync::Arc;
 use futures::{Future, Stream};
 use hyper::server::Http;
 use hyper::Chunk;
-use tokio::executor::thread_pool;
+use tokio::executor::{self, thread_pool};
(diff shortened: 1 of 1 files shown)


colinbankier:
:+1:

PR #255, opened 2018-07-15
Run Clippy from travis

im-0:
Also, this fixes few existing errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
       env:
         # Give a useful display name in the job list.
         - SHARD=rustfmt
-        - PATH=$HOME/.cargo/bin/$PATH
+        - PATH=$HOME/.cargo/bin/:$PATH
       before_script:
         - rustup component add --toolchain stable rustfmt-preview
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,16 @@ matrix:
       script:
         - echo "Checking Gotham codebase with rustfmt release `cargo fmt --version`."
         - cargo fmt --all -- --write-mode=diff
+    # Run Clippy in its own shard.
+    - rust: nightly-2018-06-29
+      env:
+        - SHARD=clippy
+        - PATH=$HOME/.cargo/bin/:$PATH
+      before_script:
+        - cargo install clippy --version 0.0.211
+      script:
+        - echo "Checking Gotham codebase with Clippy release `cargo clippy --version`."
+        - cargo clippy --all --profile test
   allow_failures:
     - rust: nightly
 addons:
--- a/gotham/src/extractor/internal.rs
+++ b/gotham/src/extractor/internal.rs
@@ -879,11 +879,11 @@ mod tests {
                 out.write_str("string")
             }
 
-            fn visit_string<E>(self, v: String) -> Result<Vec<u8>, E>
+            fn visit_str<E>(self, v: &str) -> Result<Vec<u8>, E>
             where
                 E: Error,
             {
-                Ok(v.into_bytes())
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -16,12 +16,22 @@ matrix:
...
       script:
         - echo "Checking Gotham codebase with rustfmt release `cargo fmt --version`."
         - cargo fmt --all -- --write-mode=diff
+    # Run Clippy in its own shard.
+    - rust: nightly-2018-06-29
+      env:
+        - SHARD=clippy
+        - PATH=$HOME/.cargo/bin/:$PATH
+      before_script:
+        - cargo install clippy --version 0.0.211
```
whitfin: I think it's ok to not pin a version here - we can just use latest, seeing as it's CI.

gotham/src/extractor/internal.rs
```diff
@@ -765,8 +766,8 @@ mod tests {
         assert_eq!(p.u16_val, 40511);
         assert_eq!(p.u32_val, 4000000000);
         assert_eq!(p.u64_val, 9000000000);
-        assert_eq!(p.f32_val, 1.4);
-        assert_eq!(p.f64_val, 2.6);
+        assert!((p.f32_val - 1.4).abs() < std::f32::EPSILON);
+        assert!((p.f64_val - 2.6).abs() < std::f64::EPSILON);
```
whitfin: This looks bizarre to me; what exactly is Clippy complaining about? This is a totally non-obvious thing to stumble across in the codebase.

.travis.yml
im-0: There is a problem with current Clippy: latest Clippy not always compiles with latest nightly Rust, because internal APIs used by Clippy are constantly changing.

So I think it will be better to pin specific version of nightly Rust and specific version of Clippy until Clippy will be added to the rustup as a separate component (like rustfmt).

gotham/src/extractor/internal.rs
im-0: Clippy complains that performing strict equality check on two floating point numbers obtained using different code paths is a bad idea. Even if you expect mathematically same results, results of actual floating point computations may differ because precision is not infinite.

This particular code worked before because both values was calculated using the same algorithm: conversion from string to f32/f64. First value - by Gotham code, second - by rustc in compile time. Anyway, "1.4" converted to the binary floating point format is not actually equals to the real world 1.4. Easy experiment using python:

```
>>> "%.20f" % 1.4
'1.39999999999999991118'
```

However, if you think that old code is still good enough, there is way to allow specific problems on selected blocks of code using Clippy.

.travis.yml
whitfin: @im-0 that seems reasonable, well thought out!

gotham/src/extractor/internal.rs
whitfin: It's more that I'm not entirely sure we should care (yet) about test code & Clippy. People don't write tests as it is, don't want to put them off more due to weird things like this rule.

im-0:
> I'm not sure we care about Clippy + test code at this point; it's not that critical for the time being.

It is possible to disable this, just by removing `--profile test` from Clippy command line.

> If I merge this, will all builds fail until we fix all Clippy issues?

Clippy blocks build (== returns non-zero return code) only on errors. All errors that found in current master are fixed by this pull request.

Clippy warnings are harmless, and I think I'll fix them later, when migration to hyper 0.12 will be finished and merged.

whitfin:
@im-0 ok, perfect. We'll see how it goes with the test code, we'll leave it for now!

Thanks for this PR so quickly :)

PR #261, opened 2018-07-25
Document new middleware naming conventions

secretfader:
This PR introduces the middleware naming convention that was discussed in Gitter. The idea is to create a seamless experience for newcomers who will probably start a project with core middlewares (sessions, workers, diesel, etc.), before quickly progressing to community supported middleware crates.

Introducing the `gotham-middleware-{type}` naming convention allows for middlewares to be easily grouped in search, and directly maps to the slugged URL on crates.io.

Since the cargo manifest format [supports up to 5 identifiers](https://doc.rust-lang.org/cargo/reference/manifest.html) in the `keywords` field, convention is to *also include* `gotham-middleware` as your initial keyword to further aid discoverability. The additional four keywords are entirely up to the author.

Once all documentation has been updated, then we can merge this PR.

### DIFF
--- a/middleware/template/Cargo.toml
+++ b/middleware/template/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "middleware-template"
+name = "gotham-middleware-template"
 version = "0.0.0"
 authors = ["Shaun Mangelsdorf <s.mangelsdorf@gmail.com>",
            "Bradley Beddoes <bradleybeddoes@gmail.com>"]


nicisnothere:
I've been looking for perspectives on crate name standardization from the greater community and found [this comment](https://twitter.com/nick_r_cameron/status/966753533155422208). "Nearly all crates use hyphens, so it is easier to recommend to continue to do that rather than trying to boil the ocean of making every crate change. Also hyphens are common in other ecosystems."

Should we rename all Gotham-related crates with underscores to a hyphenated format? It seems that's the long-term trend.

nyarly:
@secretfader Just pinging you about this. Conspicuous docs would be fabulous. Still game?

secretfader:
@nyarly Thanks for the po pushing this to completion. ing. I'd like to see this effort completed, but I was unclear on the Gotham community's willingness to rename existing crates to match the criteria I linked above.

Docs are useful, sure, but I think the project should also lead by example in naming our own crates. I didn't hear much response beyond the initial flurry of interest, so I sidelined the initiative. If I can get at least one clear thumbs up, I'll go ahead and rebase this PR against master and make necessary changes.

secretfader:
After more thought (and test publishes of `hypenated_lib` that I attempted to convert into `hypenated-lib`), I've decided to drop the idea unless I can get changes implemented in crates.io.

My initial understanding was that owners could change the slug, increment the version number, and republish. Crates.io, I thought, would respond by merging the hyphenated and underscored versions. I was incorrect: if we've already published with underscores, *there's currently no mechanism for renaming an existing crate*.

So I guess there isn't a change to be made, at least not until I've petitioned the crates.io team to implement the merge functionality. It seems like a fair number of authors may want this, however, I expect it would be fairly low priority even if they do add such a feature.

secretfader:
I'll start a new branch off master to make documentation changes.

PR #264, opened 2018-08-15
Re-introduce the headers helper for internal headers

### LINKED ISSUE #263: Come up with a way to frame internal headers for easy re-use
We had `XRequestId` and `XRuntimeMicroseconds` before the Hyper 0.12 update. It would be nice to get those back as strong types, instead of relying on `insert("x-request-id")` everywhere.

whitfin:
When Hyper 0.12 was adopted in #246, we removed `::gotham::helpers::http::header` as it basically broke compatibility completely. Several of the previously exposed headers are now available via the `http` crate and/or `hyper`. However Gotham still uses two of the previous headers in `x-request-id` and `x-runtime-microseconds`. 

In the adoption PR these headers are dealt with using strings directly, but this PR re-introduces the helper to make them public as `&'static str`. Whilst this is not as efficient as exposing `HeaderName` values, you can drop them directly into Hyper calls and it'll work just fine:

```rust
res.header_mut().insert(X_REQUEST_ID, "blah".parse().unwrap());
```

This is not only better for compile time checking, but it's also a little more performant as (in theory) Hyper can avoid copying the values due to them being static. I did the same for the built-in headers that are always attached to responses, too.

This would fix #263.

### DIFF
--- a/gotham/src/helpers/http/header/mod.rs
+++ b/gotham/src/helpers/http/header/mod.rs
@@ -1,0 +1,8 @@
+//! Headers recognised by Gotham which do not exist in the standard headers
+//! provided by the Hyper library.
+
+/// Marks the identifier of a request to a Gotham server.
+pub const X_REQUEST_ID: &'static str = "x-request-id";
+
+/// Marks the execution time of a Gotham request in microseconds.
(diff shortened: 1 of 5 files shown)


whitfin:
@colinbankier I'm not sure on that either, pretty odd considering it says coverage would increase.

Do we want to merge this anyway?

PR #268, opened 2018-08-18
Remove a need to allocate a Vec on response bodies

### LINKED ISSUE #244: Remove need for Vec allocation in Response body
Gotham is quite slow in benchmarks (relatively), and one of the big reasons is that the response body as it stands must be `Vec<u8>`. This means that even the simplest "hello world" cannot exist without an allocation. We should use some sort of trait to remove this allocation, to allow things like `&[u8]` and `&'static str` which would provide better ergonomics and also better performance when allocation is not needed.

whitfin:
This fixes #244.

This PR removes the need to allocate a `Vec` every time you want to return a body, rather opting for anything that Hyper can turn into a `Body`. 

The changes to signatures are designed to reduce friction for developers due to the generics involved. It is a breaking signature change, but it's easily fixable and the signature has already broken due to the Hyper 0.12 migration anyway.

I imagine this will be iterated on prior to being released, so I don't expect this to be the final set of signatures (for example, there's really no point providing a `(body, mime)` vs `body, mime`).

### DIFF
--- a/examples/cookies/introduction/src/main.rs
+++ b/examples/cookies/introduction/src/main.rs
@@ -35,10 +35,7 @@ fn handler(state: State) -> (State, Response<Body>) {
         create_response(
             &state,
             StatusCode::OK,
-            Some((
-                format!("Hello {} visitor\n", adjective).as_bytes().to_vec(),
-                mime::TEXT_PLAIN,
-            )),
+            Some((format!("Hello {} visitor\n", adjective), mime::TEXT_PLAIN)),
         )
     };
     {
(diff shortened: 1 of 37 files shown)


whitfin:
@colinbankier I think so; everything is functional and we can always iterate on it before release - as long as you and @nyarly are cool with it being merged.

nyarly:
On reflection, I'm :+1:

It opens issues in my mind wrt to conversions into `(Into<Body>, Mime)` and starts to imply that that tuple *does* want a type name, but honestly I think we can fold this into the big break-the-world and get a release out and iterate from there.

PR #274, opened 2018-09-29
Refactor default headers to be an optional middleware

### LINKED ISSUE #269: Default headers are not being correctly attached
This broke during the Hyper 0.12 adoption; the default headers (via `set_headers`) in the response module are no longer being set on responses. This should be fixed prior to v0.3 to avoid breaking the headers contract.

whitfin:
This fixes #269.

This PR will split out the default headers into a `SecurityMiddleware` that users can attach to optionally use security headers. Whilst in the past the headers were enabled by default, there is no way to disable them so I believe this makes a lot of sense.

It also removes a chunk of unnecessary functions from the helper API, instead including some more general helpers. The removed functions were a bit too low-level and prone to change to keep in the public API.

### DIFF
--- a/gotham/src/helpers/http/response/mod.rs
+++ b/gotham/src/helpers/http/response/mod.rs
@@ -1,21 +1,12 @@
 //! Helpers for HTTP response generation
 
-use http::response;
-use hyper::header::{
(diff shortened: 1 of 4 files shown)


whitfin:
Note the CI failure is due to formatting; not going to fix in this PR because it's already present in master so it would just end up being included in all PRs.

PR #275, opened 2018-09-29
Refactor test requests to remove unnecessary bindings

### LINKED ISSUE #243: Refactor from test::request::RequestBuilder to http::request::Builder
The `TestClient` relies heavily on a gotham `RequestBuilder` type that has a complete analogue in `http` but the interface is somewhat different. It would simplify our codebase and generalize the interface to replace the one with the other.

whitfin:
This fixes #243.

This PR replaces some unnecessary bindings in the test client based on requests, instead opting for using Hyper requests. `TestRequest` dereferences to `Request<Body>`, allowing you to customize the Hyper request directly. I kept the outer structure in case we want to add helpers in future (for example, `with_header` is much easier than the alternative). 

I also removed the `*_uri` functions from the client in favour of accepting `Uri: HttpTryFrom<U>` which reduces a bunch of bloat too.

### DIFF
--- a/gotham/src/handler/assets/mod.rs
+++ b/gotham/src/handler/assets/mod.rs
@@ -204,35 +204,36 @@ fn create_file_response(options: FileOptions, state: State) -> Box<HandlerFuture
 
     let (path, encoding) = check_compressed_options(&options, &headers);
 
-    let response_future = File::open(path).and_then(|file| file.metadata()).and_then(
-        move |(file, meta)| {
-            if not_modified(&meta, &headers) {
-                return Ok(http::Response::builder()
-                    .status(StatusCode::NOT_MODIFIED)
-                    .body(Body::empty())
-                    .unwrap());
-            }
-            let len = meta.len();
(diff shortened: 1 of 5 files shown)


gotham/src/handler/assets/mod.rs
```diff
@@ -518,33 +518,38 @@ mod tests {
...
+        {
+            request.headers_mut().insert(
                 IF_NONE_MATCH,
                 HeaderValue::from_bytes("bogus".as_bytes()).unwrap(),
-            ).perform()
-            .unwrap();
+            );
+        }
+
+        let response = test_client.perform(request).unwrap();
```
nyarly: In this section (and ones like it) I'm not sure I see the advantage. I can see reusing the test_client, but headers_mut().insert seems like a step back from with_header. What am I missing?

whitfin: You're not actually missing anything - originally I removed `with_header`, so I had to rewrite this. 

Then I put it back and forgot to put this back, I'll do that now :p

PR #279, opened 2018-10-02
Implement IntoResponse for more type combinations

### LINKED ISSUE #277: Add a new trait to make it easier to return responses
Right now the return type for any handler is `(State, Response<Body>)`. It would be nice in future if this were trait driven, such as `(State, Into<Response<Body>>)` to allow returning things without having to call `create_response` manually. Initially, I think it would be a good idea to allow the return of these constructs in addition to the obvious `Response<Body>`:

```rust
Into<Body>
(Mime, Into<Body>)
(StatusCode, Mime, Into<Body>)
```

Such handling would make it trivial to treat handlers as actual functions, to also make testing easier in some cases. For example, here's a potential "hello world":

```rust
fn handler(state: State) -> (State, &'static str) {
  (state, "Hello, World!")
}
```

For any other complicated responses, we can simple let `create_response` accept the same trait as whatever we define for the cases above, and then the user can handle additional headers, etc. 

This would move us a step closer to being able to return simply `-> String`, and have that be a valid response body.

whitfin:
This fixes #277, kinda.

This PR will introduce some new constructs for returning responses, namely:

```rust
Into<Body>
(Mime, Into<Body>)
(StatusCode, Mime, Into<Body>)
```

There are a few things to note:

- We can't actually implement for `Into<Body>` due to clashes with `Response<Body>`, so they're manually derived for the meantime (somewhat annoying, but not the end of the world).
- The default mime type is `TEXT_PLAIN`; not sure if it should be something else?
- The default status code is `OK`. 

Each implementation just passes through to the next, and the actual response is built using the `create_response` function in the helper - to keep things consistent. I adjusted the signature there to align with the new types as well. 

The idea is that `create_response` is now only needed when you need to do something to the response itself manually; such as changing headers, and so on. If you only care about status/mime/body, you can just return a Tuple (which should theoretically work for a lot of cases).

In doing all of this, the generic type `B` was removed from `IntoResponse` - which isn't breaking as it was introduced between `0.2.1` and now anyway. I think aligning on `Body` for the response body type makes sense at least for the time being.

Finally I removed `construct_response` in favour of making `create_empty_response` the base response creator. I doubt there's much overhead to doing it this way, and even so it makes the code more readable so likely a non-issue.

A few remaining questions I'd like opinions on:

1. Given that the point of this is to discourage `create_response` directly, should I update all examples to use these new formats?
2. Should `create_response` simply become a function that accepts `State, IntoResponse`? It can then just attach any default headers (etc.) onto the returned response.
3. Is `TEXT_PLAIN` an appropriate default mime type?

### DIFF
--- a/examples/cookies/introduction/src/main.rs
+++ b/examples/cookies/introduction/src/main.rs
@@ -31,13 +31,13 @@ fn handler(state: State) -> (State, Response<Body>) {
             .unwrap_or_else(|| "first time".to_string())
     };
 
-    let mut response = {
-        create_response(
-            &state,
-            StatusCode::OK,
-            (format!("Hello {} visitor\n", adjective), mime::TEXT_PLAIN),
-        )
-    };
+    let mut response = create_response(
+        &state,
+        StatusCode::OK,
+        mime::TEXT_PLAIN,
+        format!("Hello {} visitor\n", adjective),
+    );
+
     {
         let cookie = Cookie::build("adjective", "repeat")
             .http_only(true)
@@ -46,6 +46,7 @@ fn handler(state: State) -> (State, Response<Body>) {
             .headers_mut()
             .append(SET_COOKIE, cookie.to_string().parse().unwrap());
     }
+
     (state, response)
 }
 
--- a/examples/handlers/async_handlers/src/main.rs
+++ b/examples/handlers/async_handlers/src/main.rs
@@ -111,7 +111,7 @@ fn series_handler(mut state: State) -> Box<HandlerFuture> {
     // All we do is move `state` in, to return it, and convert any errors that we have.
     Box::new(data_future.then(move |result| match result {
         Ok(data) => {
-            let res = create_response(&state, StatusCode::OK, (data, mime::TEXT_PLAIN));
+            let res = create_response(&state, StatusCode::OK, mime::TEXT_PLAIN, data);
             println!("series length: {} finished", length);
             Ok((state, res))
         }
@@ -161,7 +161,7 @@ fn loop_handler(mut state: State) -> Box<HandlerFuture> {
 
     Box::new(data_future.then(move |result| match result {
         Ok(data) => {
-            let res = create_response(&state, StatusCode::OK, (data, mime::TEXT_PLAIN));
+            let res = create_response(&state, StatusCode::OK, mime::TEXT_PLAIN, data);
             println!("loop length: {} finished", length);
             Ok((state, res))
         }
@@ -225,7 +225,7 @@ fn parallel_handler(mut state: State) -> Box<HandlerFuture> {
 
     Box::new(data_future.then(move |result| match result {
         Ok(data) => {
-            let res = create_response(&state, StatusCode::OK, (data, mime::TEXT_PLAIN));
+            let res = create_response(&state, StatusCode::OK, mime::TEXT_PLAIN, data);
             println!("parallel length: {} finished", length);
             Ok((state, res))
         }
--- a/examples/handlers/simple_async_handlers/src/main.rs
+++ b/examples/handlers/simple_async_handlers/src/main.rs
@@ -85,7 +85,7 @@ fn sleep_handler(mut state: State) -> Box<HandlerFuture> {
     // IntoHandlerError.
     Box::new(sleep_future.then(move |result| match result {
         Ok(data) => {
-            let res = create_response(&state, StatusCode::OK, (data, mime::TEXT_PLAIN));
+            let res = create_response(&state, StatusCode::OK, mime::TEXT_PLAIN, data);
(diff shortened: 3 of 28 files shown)


whitfin:
Not to bias anything, but just to record my thoughts on the questions above at this point:

1. Updating the examples makes sense to me, given that it will make them much cleaner to read - but that might be a personal preference.
2. I think so... we can create a common `extend_response` (like existed in the past).
3. Really my argument here is just that it fits with returning `"raw strings"`.

tanriol:
I'd suggest adding plain `StatusCode`, which uses the code's canonical reason as a body.

Also, it may be a nice idea to have some typed wrappers to be able to write just `typed::Json(json_string)`.

nyarly:
1. Yeah, I'd say update examples to match the intended use.
2. This seems like a bigger question. Separate comment.
3. I think it's pretty reasonable, especially if...

@tanriol has some good ideas there, too. If it seems like a hassle, I'm happy to seem them promoted to an issue - maybe something I can pick up near term. StatusCode, (StatusCode, Into<Body>), etc. Tuple variants with headers seem reasonable as well.

There's value to being able to easily specify the MIME type, but (just to completely bikeshed this), is `typed::Json(data)` better than `(mime::JSON, data)`?

whitfin:
I agree with updating examples... I'll get to that today or tomorrow with luck.

@tanriol for returning just `StatusCode`, I'm all for this but you're saying the body would just be empty? If so, I don't actually see much use case for this.

@nyarly FWIW I agree that if we do move on that it should be taken up at a later date, I'm not convinced that `(mime::JSON, data)` is any worse than `typed::Json(data)` - and if you wanted, you could always roll this in your code quickly enough.

nyarly:
Bikeshedding intensifies:

Handlers, for the forseeable future, won't be able to return impl IntoResponse, I think. At least, not methods-as-handlers. And I'm not sure about how https://github.com/gotham-rs/gotham/blob/master/gotham/src/service/trap.rs#L42 interacts with `impl` returns. Right?

Granted that, we'll need some way to return (State, Response) instead of (State, impl IntoResponse), and `create_response(s: State, r: impl IntoReponse) -> Reponse<Body>` would be one way to do that.

I'm leery a little of changing the interface that much. Client code needs a bunch of updates, etc, but we're still 0.0, so you take you chances, I suppose. I agree with the existing signature change, though, so :man_shrugging: 

It occurs to me though: we already have a function like `(s: State, r: impl IntoReponse) -> Reponse<Body>` - `into_response()`.

whitfin:
@nyarly yeah, you still need to `(State, IntoResponse)` until we figure out a way to make `State` a reference.

As it stands, just so I understand, you can return `(State, Response<Body>)` because `Response<Body>` implements `IntoResponse` - so `(State, Response<Body>)` is the same as `(State, IntoResponse)`. 

The proposed change to `create_response` is basically so that you can do the conversion the trait does on the handler, but then modify the response as you like before actually returning. It is largely breaking, but it's also easier to avoid breaking in the future given that it's a trait going forward. So I'm not really sure.

You're right about the redundancy though, `create_response` and `into_response` are identical - removing one just felt a little off because each allows different code style.

tanriol:
@nyarly The point with plain `StatusCode` would be to have an obvious way to return "bare" errors such as `StatusCode::BAD_REQUEST` and especially `StatusCode::INTERNAL_SERVER_ERROR` (normally I wouldn't want to provide any custom explanation to the client for this one, would you?).

nyarly:
Yeah, I'm with @tanriol on those - there should be standard (like RFC 723x) messages to put in the body for those. I'd hope that `http` would include them, even.

whitfin:
I'm still not sure I grasp; I get the concept of returning a `StatusCode`, sure, but I think it would have to always be an empty body - otherwise you have to start respecting content types and what not.

e.g. if I return `StatusCode::INTERNAL_SERVER_ERROR`, it should just throw a 500 to the client without a body.

nyarly:
It sounds like we don't agree yet on what that would do. I'd've thought
that `StatusCode::INTERNAL_SERVER_ERROR` would become a 500, Content-Type:
text/plain, Body: "Internal server error"

whitfin:
I mean if there's a spec for what to return, then it's fine - but I just don't want us to invent bodies for the codes, if that makes sense?

whitfin:
Regardless of what we decide on the StatusCode, should we merge this as-is? I wouldn't consider StatusCode as a blocking item here (although feel free to disagree!).

nyarly:
I tend to agree. I thought I'd suggested that the additional changes made sense as new issues?

I'm :+1: as soon as we have CI passing.

whitfin:
@nyarly you got it - CI should be good now!

PR #283, opened 2018-10-23
Add static assets example to readme.

colinbankier:
I noticed the static assets example missed being added to the examples readme.


whitfin:
Nice catch :p

Issue #111, opened 2018-01-23
Create multi pipeline example


bradleybeddoes:

colinbankier:
I'm happy to take a stab at this - but I'm not sure exactly what was initially intended.
There are some examples of multi pipelines from https://github.com/gotham-rs/gotham/pull/191 and https://github.com/gotham-rs/gotham/issues/190 and https://github.com/gotham-rs/gotham/issues/258

I imagine just something similar, but in a stand-alone example.
I'm thinking an example showing different pipelines for serving html and a json api could be a common use-case. @bradleybeddoes - anything else specific you intended to demonstrate here?

msrd0:
I believe this was done in #287

PR #287, opened 2018-11-07
Add multiple pipeline example.

colinbankier:
This adds an example of using multiple pipelines for https://github.com/gotham-rs/gotham/issues/111.

I attempted to create a "believable" scenario showing multiple routes so that others could see how to implement their own use-cases, and show different ways to override middleware for different routes.

Hopefully the example isn't too long-winded because of this.
Such an example certainly highlights that the API for setting up a pipeline set is pretty cumbersome...something I think could be reviewed in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 target
 Cargo.lock
 .idea/
-rls
+
 *.rustfmt
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 target
 Cargo.lock
 .idea/
-
+rls
 *.rustfmt
--- a/examples/middleware/multiple_pipelines/src/main.rs
+++ b/examples/middleware/multiple_pipelines/src/main.rs
@@ -1,0 +1,270 @@
+//! Provides examples of using multiple middleware pipelines for different routes.
+//!
+//! We'll create an app with a few different routes that need
+//! different combinations of middleware.
+//!
+//! By default, we'll expect users to be logged in, using a session middleware
+//! to track client cookies.
+//! On the homepage, however, we don't want users logged in, so we'll override
+//! this path to not use any middleware.
+//! Our app also has some admin functionality, so we'll also override admin paths
+//! to require an additional admin session.
+//! Finally, our app exposes an JSON endpoint, which needs its own middleware.
+extern crate futures;
+extern crate gotham;
+#[macro_use]
+extern crate gotham_derive;
+
+#[macro_use]
+extern crate serde_derive;
+extern crate hyper;
+extern crate mime;
+
+use futures::future::{self, Either};
+use hyper::header::{HeaderMap, ACCEPT};
+use hyper::{Body, Response, StatusCode};
+
+use gotham::handler::HandlerFuture;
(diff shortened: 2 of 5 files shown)


secretfader:
I'd like to see this merged, provided that it builds when rebased on `master`. Multiple pipelines are a feature I use often, but still largely misunderstood.

.gitignore
```diff
@@ -1,5 +1,5 @@
 target
 Cargo.lock
 .idea/
-
+rls
```
whitfin: Not a big deal, but I'm curious what caused this to be added?

examples/middleware/multiple_pipelines/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn call<Chain>(self, state: State, chain: Chain) -> Box<HandlerFuture>
+    where
+        Chain: FnOnce(State) -> Box<HandlerFuture> + 'static,
+    {
+        let accepts = HeaderMap::borrow_from(&state)
+            .get(ACCEPT)
+            .map(|ct| ct.to_str().unwrap().to_string());
+
+        let f = match accepts {
+            None => Either::A(chain(state)),
```
whitfin: I think you could technically avoid the `Either` if you just return `Box::new(chain(state))` from this branch, and a Box directly in the other.

Not a big deal, but might be less noisy.

colinbankier:
Great, thanks. I'll take a look at updating this and getting it cleaned up.

examples/middleware/multiple_pipelines/src/main.rs
colinbankier: Thanks - it was a while ago - but I prob got confused about mismatching future types.
Good to know this is the better solution.

.gitignore
colinbankier: Yeah, I'll clean this up. Was probably some local config I had for rls dumping logs there - I don't think it's useful or valid anymore.

examples/middleware/multiple_pipelines/src/main.rs
```diff
@@ -73,7 +73,7 @@ impl Middleware for ApiMiddleware {
                     mime::APPLICATION_JSON,
                     body,
                 );
-                Either::B(future::ok((state, response)))
+                Box::new(future::ok((state, response)))
```
secretfader: Is it possible to pass `future::ok((state, response))` here instead of `Box`? I believe `f` is double-boxed on line 79.

secretfader:
I can't wait to have a multiple pipeline example in the repo. This is something I wrestled with during my early days using Gotham, and I suspect others have too. Nice work on a "real world" example, @colinbankier!

whitfin:
This looks good, don't see anything that holds this back!

examples/middleware/multiple_pipelines/src/main.rs
colinbankier: Thanks @secretfader - I missed that.
The `match` has to return a box since `chain` returns a box in the other arms, but you're right, I had missed the extra box at the end. Opened #306 to tidy that up, since this is merged :)

PR #290, opened 2018-11-25
Add ability to spawn futures ontop of the TestServer

LucioFranco:
For https://github.com/toshi-search/Toshi we need to spawn some additional futures along side the test server to be able to properly test some features. This change exposes the spawn method from the internal test server runtime.

### DIFF
--- a/gotham/src/test/mod.rs
+++ b/gotham/src/test/mod.rs
@@ -115,6 +115,20 @@ impl TestServer {
         self.client_with_address(SocketAddr::new(IpAddr::from([127, 0, 0, 1]), 10000))
     }
 
+    /// Spawns the given future on the `TestServer`'s internal runtime.
+    /// This allows you to spawn more futures ontop of the `TestServer` in your
+    /// tests.
+    pub fn spawn<F>(&self, fut: F)
+    where
+        F: Future<Item = (), Error = ()> + Send + 'static,
+    {
+        self.data
+            .runtime
+            .write()
+            .expect("unable to acquire read lock")
+            .spawn(fut);
+    }
+
     /// Returns a client connected to the `TestServer`. The transport is handled internally, and
     /// the server will see `client_addr` as the source address for the connection. The
     /// `client_addr` can be any valid `SocketAddr`, and need not be contactable.
--- a/gotham/src/test/mod.rs
+++ b/gotham/src/test/mod.rs
@@ -191,15 +191,9 @@ impl TestServer {
         R: Send + 'static,
         E: failure::Fail,
     {
-        self.data
-            .runtime
-            .write()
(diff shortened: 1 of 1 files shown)


colinbankier:
Thanks @LucioFranco, happy to merge if this is useful to you :+1: 
Since this becomes part of the `TestServer` public api, my only thought was that it'd be nice if it had a test. 
A convenient way to do that didn't jump out me however - so if you have a way, that'd be great, otherwise, I think we can just merge it in.

LucioFranco:
@colinbankier I agree a test would be nice, but honestly, it just calls out to tokio so hopefully spawn is tested there. There really isn't any extra logic being applied that would be useful to test.

whitfin:
+1 on it, seems good. Minor nitpick; can you please make the private `run_future` in that file re-use this public API? That should address @colinbankier's concerns abouta test (seeing as the server will then be using the new API).

LucioFranco:
@whitfin if you're talking about this [run_future](https://github.com/gotham-rs/gotham/blob/42d923b2e3c3bb6fd27683a1f56fb52528aef1d9/gotham/src/test/mod.rs#L188) it is using `block_on` and not `spawn`. So changing it to use `spawn` instead of `block_on` will change how the tests work, since `spawn` doesn't block till completion. Maybe I am looking at the wrong `run_future`?

whitfin:
@LucioFranco you're right! But you can make a fake blocking implementation like this:

```rust
 let (p, c) = oneshot::channel::<Result<R>>();

let f = future.map_err(|err| err.into()).then(|res| {
    if let Err(_) = p.send(res) {
        warn!("unable to respond with result");
    }
    Ok(())
});

self.spawn(f);

c.wait().unwrap()
```

This would re-use the public spawn, but block until the result is provided.

LucioFranco:
@whitfin That would work but don't know if that is worth it to be honest, why not just use block_on? Also with that implementation you throw away the error as a warning that will get swallowed via `cargo test`, wouldn't it be better possibly panic, so at least the error is captured.

whitfin:
@LucioFranco as above; we either need a test (because it's part of the public API), or we can just re-use it internally from `run_future` because then it's tested anyway (since `run_future` backs many existing tests).

Sure, panic is likely better - it's sample code :) But yes, the point here is to get test coverage of the new public API.

LucioFranco:
@whitfin Sounds good to me. I've shamelessly ripped out what `block_on` does within `tokio::runtime::Runtime` minus dealing with the executor. The unwrap should panic the result from the future.

whitfin:
@LucioFranco awesome, looks great to me! I'll leave merging up to @colinbankier since he went through it initially, but thanks for the contribution!

PR #293, opened 2018-12-04
Add a helper to use pipelines with (New)Handlers

### LINKED ISSUE #292: Middleware running before router?
Hi, I've recently started experimenting with Gotham and so far have to say Thank You, for an awesome library! Now to my issue:

If I understand my debug logs correctly, any middleware I add to a `Router` will only run if the `Router` successfully matches a route. I have a middleware<b><sup>1</sup></b> I want to run irrespective of whether the `Router` has any matching routes (or rather, I want the middleware to run _before_ the `Router`).

Is there a standard way of doing this?

I haven't had the time yet to experiment, but I've thought of the following approaches:

* "Manually" wrap the middleware around the `Router` by implementing `(New)Handler`.
* Use `PipelineHandleChain` to do the same.
* Use two nested `Router`s: The outer one has the middleware I want to run and a `*` delegation to the inner router (not sure if that would work).

<b><sup>1</sup></b>: The actual use case is that I want to reject all unauthenticated requests with a 403. If the request is authenticated, I want the normal 404 handling of the router to take place. If Gotham has other ways to deal with this, I'd be happy to hear about them.

TimNN:
_(The code is still missing tests & documentation, but I would like to get some initial feedback before I add those)._

This was the solution I ended up with for #292. While cleaning the code up, I remembered that when I was first looking into solutions for my issue, I was very surprised that there didn't seem to be a standard way to use pipelines with plain handlers, only `Routers` appeared to be supported. So I decided to phrase my solution in those terms: A helper to use a pipeline with a plain handler. _(I'd be happy to accept suggestions regarding naming, code organization and better bounds)._

_(Also note that I find the behavior of the `Router`, to only apply pipelines when it successfully matches a route, unexpected and surprising. But even if that were to change, I think the utility proposed here would still be valuable since it allows using pipelines without the router)._

cc @whitfin, @colinbankier 

Fixes #292

### DIFF
--- a/gotham/src/pipeline/chain.rs
+++ b/gotham/src/pipeline/chain.rs
@@ -5,7 +5,8 @@ use borrow_bag::{Handle, Lookup};
 use futures::future;
 use std::panic::RefUnwindSafe;
 
-use handler::{HandlerFuture, IntoHandlerError};
+use error::Result;
+use handler::{Handler, HandlerFuture, IntoHandlerError, NewHandler};
 use middleware::chain::NewMiddlewareChain;
 use pipeline::set::PipelineSet;
 use pipeline::Pipeline;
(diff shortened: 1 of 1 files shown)


whitfin:
> Also note that I find the behavior of the Router, to only apply pipelines when it successfully matches a route, unexpected and surprising.

I used to think this when I initially got involved with this project, but have since come to realise that it's much better this way. You don't want to run something like authentication middleware on 404s; it'd be a complete waste of time and energy. If you /do/ want that, it's much better to explicitly do it some other way (which is currently missing, afaik).

I believe all that would happen is that people would start getting confused why things are firing when the routes are invalid. There's no real answer; different frameworks do it differently. We should likely just add some sort of configuration to enable middlewares to fire on 404 (and default to off, for compatibility), and allow the developer to choose. 

I'm unsure if there's a way to adapt the existing API space to cover this.

nyarly:
I'm late to review this, and for that I apologize.

If you're looking to have a pipeline added to every route, that's what the first argument of [gotham::router::build::build_router()](https://docs.rs/gotham/0.3.0/gotham/router/builder/fn.build_router.html) is for. Am I missing some requirement that isn't served by that function?

TimNN:
It's been a while since I looked at this, but IIRC the issue I had back then was that `build_router` only applies the middleware if it successfully matches a route.

nyarly:
Got it. So the idea was to add a middle-applying handler that you could wrap around the router-as-Handler. That makes sense.

What confused me is that this approach allows for wrapping *any* Handler in a pipeline, which means that there'd be two ways for a pipeline to be interposed above the resource handler: either in its route, or with a wrapper. That strikes me as a recipe for confusing behaviors and gnarly bugs.

I don't want to shut out your use case, and I apologize again for jumping in late here.

As I see it, we could use pipeline wrappers on handlers everywhere, or specialize this wrapper to surrounding Routers.

What do you think, @colinbankier @whitfin @TimNN ?

nyarly:
@TimNN Apologies for the delays here. Are you inclined to resolve the conflicts here and land it?

msrd0:
I'm closing this PR as it has not seen any activity in the last 2 years.

PR #294, opened 2018-12-08
Add askama templating example

rakenodiax:
I've added an example of using the `askama` templating engine. This example borrows heavily from the `tera` templating example, but also uses some common patterns specific to `askama`, such as a build file to recompile templates when they change and template variables.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ members = [
     # Tera template
     "examples/templating/tera",
 
+    # Askama template
+    "examples/templating/askama",
+
     # routing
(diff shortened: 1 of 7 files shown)


rakenodiax:
@whitfin Done! Thank you for pointing that out.

I noticed that the Travis `rustfmt` job failed, but the diffs don't seem to be affected by my commits; should I be applying `rustfmt` on the entire codebase?

whitfin:
No, you can ignore the `rustfmt` stuff. Latest stable Rust changed the rules and I haven't run a pass on the repo yet.

Issue #289, opened 2018-11-09
JWT Middleware in Core


secretfader:
I've been developing an authentication middleware for Gotham 0.3 that [extracts, parses, and validates JSON Web Tokens when provided via the `Authorization` header](/uptimeventures/gotham-middleware-jwt).

Since this seems like pretty standard tooling for most web apps and APIs, I'd like to propose moving my crate to this repository. What do you think? Does it belong in the core repository, or just in the @gotham-rs org? Or is it better left at the current org? 

I'm open to suggestions, including amending the license (if need be). The bottom line is I'd like to place the code in a more community-friendly place, where it might pick up other maintainers who can step in when I'm out.

colinbankier:
Thanks for sharing your middleware @secretfader!  I agree we certainly want to make it easy for common tooling to be discovered and maintained, and to encourage the Gotham ecosystem to grow. I'll wait for @whitfin and @nyarly to weigh in on whether they think moving this into core is the right move.

secretfader:
Sounds good, @colinbankier. The project recently received an update enabling support for Gotham 0.3, and JWT 5.x, so all dependencies are currently up-to-date.

My initial thought was this middleware *didn't belong* in this repository, but then I realized authentication is roughly on the level of the Diesel middleware (in terms of common development practices and conveniences that speed building an app).

I'll keep an eye on this issue to see what is decided.

secretfader:
This issue has been open for a couple months. @nyarly, @whitfin, and @colinbankier, if any of you want to continue the discussion, feel free to re-open. Otherwise, I'll close it to keep your queue of work tidy.

colinbankier:
Thanks @secretfader. Yeah, I acknowledge activity has been pretty slow over the holiday period. Sorry about that.
I for one would vote for bringing this in, as JWT auth seems like a pretty common piece of functionality.
Feel free to open a PR with it :)

secretfader:
I'll send you a PR later on then, and we can discuss how to allow releases by the Gotham Core Team.

PR #296, opened 2019-01-08
Import JWT Authentication middleware

secretfader:
This PR imports the JWT Authentication middleware I was chatting about earlier in #289. The current version is set up to work with Gotham 0.3.

### DIFF
--- a/middleware/jwt/Cargo.toml
+++ b/middleware/jwt/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "gotham_middleware_jwt"
+version = "0.3.1"
+authors = ["Nicholas Young <nyoung@uptime.ventures>",
+          "Colin Bankier <colinbankier@gmail.com>",
+          "Isaac Whitfield <iw@whitfin.io>",
+          "Judson Lester <nyarly@gmail.com>",
+          "Bradley Beddoes <bradleybeddoes@gmail.com>"]
+description = "JWT middleware for the Gotham web framework."
+repository = "https://github.com/gotham-rs/gotham"
+keywords = ["gotham-middleware", "jwt", "jsonwebtoken", "authentication"]
+homepage = "https://gotham.rs"
+readme = "README.md"
+license = "MIT/Apache-2.0"
+edition = "2018"
+
+[dependencies]
+futures = "0.1"
+gotham = "0.3"
+gotham_derive = "0.3"
+serde = "1.0"
+serde_derive = "1.0"
+hyper = "0.12"
+jsonwebtoken = "5.0"
+log = "0.4"
--- a/middleware/jwt/src/middleware.rs
+++ b/middleware/jwt/src/middleware.rs
@@ -1,0 +1,328 @@
+use crate::state_data::AuthorizationToken;
+use futures::{future, Future};
+use gotham::{
+    handler::HandlerFuture,
+    helpers::http::response::create_empty_response,
+    middleware::{Middleware, NewMiddleware},
+    state::{request_id, FromState, State},
+};
+use hyper::{
+    header::{HeaderMap, AUTHORIZATION},
+    StatusCode,
+};
+use jsonwebtoken::{decode, Validation};
+use serde::de::Deserialize;
+use std::{io, marker::PhantomData, panic::RefUnwindSafe};
+
+/// This middleware verifies that JSON Web Token
+/// credentials, provided via the HTTP `Authorization`
+/// header, are extracted, parsed, and validated
+/// according to best practices before passing control
+/// to middleware beneath this middleware for a given
+/// mount point.
+///
+/// Requests that lack the `Authorization` header are
+/// returned with the Status Code `400: Bad Request`.
+/// Tokens that fail validation cause the middleware
+/// to return Status Code `401: Unauthorized`.
+///
+/// Example:
+/// ```rust
+/// extern crate futures;
+/// extern crate gotham;
+/// extern crate gotham_middleware_jwt;
+/// extern crate hyper;
+/// extern crate serde;
+/// #[macro_use]
+/// extern crate serde_derive;
+///
+/// use futures::future;
+/// use gotham::{
+///     helpers::http::response::create_empty_response,
+///     handler::HandlerFuture,
+///     pipeline::{
+///         new_pipeline,
+///         set::{finalize_pipeline_set, new_pipeline_set},
+///     },
+///     router::{builder::*, Router},
+///     state::{State, FromState},
+/// };
+/// use gotham_middleware_jwt::{JWTMiddleware, AuthorizationToken};
+/// use hyper::{Response, StatusCode};
+///
+/// #[derive(Deserialize, Debug)]
+/// struct Claims {
+///     sub: String,
+///     exp: usize,
+/// }
+///
+/// fn handler(state: State) -> Box<HandlerFuture> {
+///     {
+///         let token = AuthorizationToken::<Claims>::borrow_from(&state);
+///         // token -> TokenData
+///     }
+///     let res = create_empty_response(&state, StatusCode::OK);
+///     Box::new(future::ok((state, res)))
+/// }
+///
+/// fn router() -> Router {
+///     let pipelines = new_pipeline_set();
+///     let (pipelines, defaults) = pipelines.add(
+///         new_pipeline()
+///             .add(JWTMiddleware::<Claims>::new("secret".as_ref()))
+///             .build(),
+///     );
+///     let default_chain = (defaults, ());
+///     let pipeline_set = finalize_pipeline_set(pipelines);
+///     build_router(default_chain, pipeline_set, |route| {
+///         route.get("/").to(handler);
+///     })
+/// }
+///
+/// # fn main() {
+/// #    let _ = router();
+/// # }
+/// ```
+pub struct JWTMiddleware<T> {
+    secret: &'static str,
+    validation: Validation,
+    claims: PhantomData<T>,
+}
+
+impl<T> JWTMiddleware<T>
+where
+    T: for<'de> Deserialize<'de> + Send + Sync,
+{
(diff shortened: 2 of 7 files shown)


secretfader:
Here you go, @colinbankier. I hope everyone else approves. :+1:

colinbankier:
Great, thanks @secretfader. At a quick first scan this looks great.
There does seem to be some issue with the Windows build in one of the deps: 

> "crypto/fipsmodule/aes/aes.c"]: The system cannot find the file specified. 

I haven't looked in detail what can be done about that.

@nyarly @whitfin - any issues with including this once all checks pass?

secretfader:
I'm not sure what's going on with the Windows `-gnu` build, but similar build errors have shown up in ring [before](https://github.com/briansmith/ring/issues/330). Unless someone else with existing knowledge swings in with a tidy solution, I'll do some research and see how we can get it building on both toolchains.

secretfader:
As I read reports from around the Rust community, it seems the most common recommendation is to rely on the Windows MSVC toolchain primarily because its ABI produces better interop support for standard Windows applications. [Rustup's documentation addresses this, even.](https://github.com/rust-lang/rustup.rs)

My question then is: whether to allow failures on what is considered a lower-tier support level for most of the Rust community?

colinbankier:
I'm totally fine with switching to the MSVC toolchain if that fixes the issue easily, unless there's a good reason not to. I'd be nice to understand a little more what the issue is before switching though (if it doesn't become a rabbit hole digging into it).

whitfin:
@colinbankier as a general, I don't mind liberally merging things like middleware which are self contained and don't affect the general codebase.

However... I'm not sure if we want to take ownership of things like JWT rather than having it separately owned. I don't know enough about JWT to know if this implementation covers the entire surface area of JWT auth - I'm thinking about the case where something is missing and @secretfader doesn't have time to update it and we're left with an incomplete implementation. I guess I'm saying that I'm not really the right person to give a sign off on this :p

It might be fine just on the premise that this will be a separate crate rather than shipping with core. I know @nyarly had some thoughts on middleware distribution a while ago so I'm interested in hearing from them. 

Aside from the above, I'll give the code itself a quick pass anyway!

middleware/jwt/src/middleware.rs
```diff
@@ -0,0 +1,333 @@
...
+        trace!("[{}] pre-chain authentication", request_id(&state));
+
+        let token: Option<String> = {
+            let header = HeaderMap::borrow_from(&state).get(AUTHORIZATION);
+
+            match header {
+                Some(h) => match h.to_str() {
+                    Ok(hx) => {
+                        let parts: Vec<&str> = hx.rsplit(": ").collect();
+                        Some(parts[0].to_owned())
```
whitfin: It seems like we can avoid the `collect` here.

From what I can tell, this is basically doing one of those things where we split `Authorization: TOKEN`. If that's the case, can't we simply skip the first N bytes of the input and just reference the TOKEN directly? That makes this avoid allocations completely.

middleware/jwt/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+          "Bradley Beddoes <bradleybeddoes@gmail.com>"]
+description = "JWT middleware for the Gotham web framework."
+repository = "https://github.com/gotham-rs/gotham"
+keywords = ["gotham-middleware", "jwt", "jsonwebtoken", "authentication"]
+homepage = "https://gotham.rs"
+readme = "README.md"
+license = "MIT/Apache-2.0"
+
+[dependencies]
+futures = "0.1.25"
```
whitfin: Can you pin these versions only to the minor? i.e. `futures = 0.1`?

middleware/jwt/src/state_data.rs
```diff
@@ -0,0 +1,21 @@
+pub use jsonwebtoken::TokenData;
+
+/// Struct to contain the JSON Web Token on a per-request basis.
+#[derive(StateData)]
+pub struct AuthorizationToken<T>
+where
+    T: Send + 'static,
+{
+    /// Headers, Claims, and other data from the token.
+    pub token: TokenData<T>,
```
whitfin: I don't think this field needs to be public, we can direct people to `new` instead.

colinbankier:
Thanks for the feedback @whitfin.
I must admit, I initially glossed over the fact that it would be published as a separate crate and not ship with core. Is that the normal way we would deal with middleware that lives in this repo? I also don't want it to add unnecessary friction around release - either to gotham or the middleware crate.

One motivator for including it was simply for discoverability. There are other ways to help this though, and might not be a good reason.

secretfader:
Maybe the solution isn't, as I proposed, moving middleware into core. This thread has be wondering if I should've suggested a *document of known middlewares* or a link to the `gotham-middleware`  [keyword search](https://crates.io/keywords/gotham-middleware) on crates.io. Maybe middleware *doesn't belong in the core repo at all*.

If there's serious interest in moving this PR forward, then I'll fix the issues raised by @whitfin and we can proceed to merge at will. (In fact, most of the [changes are already made over at the old repo](https://github.com/uptimeventures/gotham-middleware-jwt), and just need to be synced to this PR.)

But I have the distinct hunch that going through this process has raised better ideas than the original. I'd hate to waste your time if there are better approaches on the table.

whitfin:
I _think_ that we are yet to determine exactly what we're doing with middleware.

The current state of things is that "general" middleware (i.e. loggers and such) as in core itself, and ship in the `gotham` crate. They're used by enough developers that they make sense to be included. 

There is a Diesel based middleware which lives in this repo, but ships as a separate middleware. This predates me on this project, but I believe it was done this way to enable iteration separately. That makes sense to me (because you can fix bugs in that without shipping the entire codebase).

However! A lot of people seem to enjoy when things "just work", so I've been thinking about a model where we basically carry on as we're currently doing (and as such, this would move into the core repository in the same way the Diesel one exists). We could them add dependencies to them and re-expose them from the `gotham` crate, perhaps via feature flags to avoid compiling them unnecessarily. This would make adoption of these middlewares much easier.

The only downside is what I said before; there's a notion that because it lives in this repository, it is owned by this team. I don't mind this (especially enough to block merging this PR), but it was mainly just a note for the future. 

So really, what I'm saying is pretty much summed up as: merge away! If we decide something later we can always change it up (there are middlewares in three places already, so we'd have to change something anyway).

@secretfader I'm very unfamiliar with JWT, but could this implementation be considered "complete"? Is there some validation document we can compare against? JWT is so general that this might actually be worth moving into the core crate itself, but I'm unsure exactly what is in this PR vs. what could be in this PR. If this is complete, I'd vote for moving it into the `gotham` crate. If it's not, then I'd have it where you currently have it targeted - which would meant it shipped under its own name.

secretfader:
This middleware is only a portion of your authentication system where JWTs are concerned, but  to answer your primary question, yes, I believe it represents the most secure approach. By default, it adopts the default JWT validation [from the `jsonwebtoken` crate](https://github.com/keats/jsonwebtoken), but allows it to be overridden with custom validation, making the validation of incoming requests even more secure.

While it will only allow tokens that validate to pass (checking that both the crypto hasn't been tampered with before checking validity of the issuer and time of any expiration keys), you'll still need to implement logic for generating and storing new JWTs, purging expired tokens, and linking them to accounts inside your system. But this module provides assurances that when you look up a user from the parsed values inside the token, those values haven't been tampered with during transport.

I have tinkered with the idea of implementing a turn-key auth platform on top of Gotham, and this crate would be a vital piece whether in core or not. or what it's worth, I think the module should probably always be distributed as a separate crate even if it lives in this repo. While authentication is a general need among web apps, not all require it (and even fewer require a specific implementation like JWT)

That's the main story, @whitfin and @colinbankier. I'd be happy to finish this up and get it ready to go if that's what you want.

middleware/jwt/src/state_data.rs
secretfader: This field has been updated in the new version, pushed just before this comment.

middleware/jwt/Cargo.toml
secretfader: Done.

middleware/jwt/src/middleware.rs
secretfader: Allocation was dropped in the latest version I pushed. Good point.

colinbankier:
Thanks for updating things @secretfader. I think, based all the discussion above, we're pretty much ready to merge it :+1: 
The last pesky thing is to get the build to pass.
@secretfader, I'm not super familiar with appveyor and the windows builds, but from a glance it seems like it should be possible in `.appveyor.yml` to add an `if` check in the `test_script` based on `TARGET` - so that we can do `cargo test -all` on x86_64-pc-windows-msvc toolchain, but add an `--exclude` for `middleware/jwt` on `x86_64-pc-windows-gnu`.

It'd be nice to allow the rest of Gotham to still build on `x86_64-pc-windows-gnu` if something like that is possible. If that doesn't work out, I'm OK with removing `x86_64-pc-windows-gnu` as a target. I'm not sure how important that is not to Windows users, but noted the recommendation you linked above for ring that "People should just use the -msvc toolchain".

secretfader:
@colinbankier I've read over the AppVeyor documentation, and it seems relatively easy to allow failures on certain targets, but that feels overly broad. The original plan of mine was to take a different approach and set different `test_script` values on a per-target basis, but I'm not sure it's possible.

secretfader:
Ok, after several minutes wrestling with AppVeyor configuration, I didn't find a way to successfully exclude crates to build/test only on a single `TARGET`.

I'd like to find a way to re-enable `x86_64-pc-windows-gnu`, but I'm not sure how to proceed even after spending more time than I should've in the docs.

colinbankier:
Thanks @secretfader - I'll have poke at it too when I can, and if not successful either, will go with what you've got here.

secretfader:
Thanks for handling this, @colinbankier. I also appreciated the helpful review, @whitfin.

Here's to making Gotham even more awesome. :+1:

PR #297, opened 2019-01-17
Update rand 0.5 to 0.6

secretfader:
The `rand` crate recently split code pertaining to the ChaCha implementation into its own crate. This PR changes Gotham to include that new dependency, and updates the core `rand` crate to version `0.6`.


secretfader:
Updated the commit message to better align with repository standards.

colinbankier:
Mmm. That is rather a lot of warnings from clippy.
I haven't been following clippy closely, but it seems related to this build picking up a newer rust version - 1.32.
Although most of the lints look like good suggestions, it seems a bit harsh to block this PR, and we should instead handle those in a separate issue. I'm having a look to see what a sensible level of lints to disable is.

colinbankier:
The only solution I can think of for now is to disable clippy most clippy lints and open a separate issue to go through and address them to re-enable it.
You can edit the clippy command in the travis config to
`cargo clippy --all --profile test -- -A clippy::all`
Then fix the one remaining issue, removing some redundant lines
```
diff --git a/misc/borrow_bag/src/lib.rs b/misc/borrow_bag/src/lib.rs
index e1f5a32..4729f76 100644
--- a/misc/borrow_bag/src/lib.rs
+++ b/misc/borrow_bag/src/lib.rs
@@ -90,9 +90,6 @@ impl<V> BorrowBag<V> {
     /// #
     /// # let _ = (bag, handle);
     /// ```
-    // This isn't add like +..
-    // Consider renaming this method?
-    #[cfg_attr(feature = "cargo-clippy", allow(should_implement_trait))]
     pub fn add<T>(self, t: T) -> (BorrowBag<V::Output>, Handle<T, V::Navigator>)
     where
         V: Append<T>,

```

secretfader:
@colinbankier I have a branch that resolves clippy lint errors in Gotham. I don't recall if I've pushed it to GitHub yet (and I'm currently mobile), but it should appear later today.

whitfin:
Yeah, I wouldn't be too worried about Clippy. It should be set to not fail Travis builds; the rules change under our feet and cause builds to fail when they did previously. I'll take a look this weekend if I get a sec!

secretfader:
I'm now back in front of the computer and will push up my Clippy resolutions momentarily. I'd like to keep dependencies current, so when this built I pushed it.

secretfader:
To my knowledge, no new Clippy errors were introduced in this PR, so it should be a clean and easy merge.

colinbankier:
Oh. Github won't let me merge. I could do it manually - or @whitfin maybe you have privileges to override that? Or should @secretfader disable the clippy check in his branch?

secretfader:
I added a commit filtering Clippy lints, @colinbankier. I'll patch up Clippy config in my other PR.

PR #298, opened 2019-01-19
Fix Clippy errors on core packages

secretfader:
This PR is where I'm attempting to fix Clippy errors that are occuring on core packages (`gotham_derive` and `borrow_bag` are initial targets).

A lot of these commits, I expect, will boil down to incoming API changes brought about by the new edition of Rust, and simplifications that came with it.


secretfader:
Here are the changes I mentioned in #297. I wrote the code yesterday and forgot to create a PR.

secretfader:
While I expect this PR will build as of now, I still have several more stylistic changes incoming.

secretfader:
I need to make these changes more readable. Maybe I can roll them up in a handful of commits, grouped by module and the type of changes made (reducing allocations, replacing duplicate type conversions, etc).

secretfader:
I'm hoping that I'll be able to push this to completion tomorrow (Monday, Jan. 21). The goal is to remove some of the most persistently frustrating lints (like `should_implement_trait`, which is fussing about the Gotham `borrow_*` APIs mostly. My take is: yes, we should probably rename those, but now isn't the most critical time.)

Issues with other lints should be resolved when I submit this for review. The goal was to bending to Clippy's will, so long as the public API doesn't change.

PR #299, opened 2019-01-21
Fix and/or reconfigure Clippy lints

secretfader:
This should be ready to go, @whitfin. Apologies for the separate PR, but the content is the same as #298. (GitHub, for some reason, wouldn't let me force update my branch without creating a new one.)

The goal was to align with Clippy's recommendations, so long as the public API didn't change. Lints that would require API or significant code changes aren't included here. Instead, lints related to changes I wasn't willing to make just yet have been explicitly allowed.

Many of these changes boil down to using `char`s over `str`, reducing allocations, and tidying up duplicate type conversions. Test coverage shouldn't change significantly, and all of the current suite should pass.


secretfader:
I'll reconfigure Clippy lints that are currently allowed to pass, and discuss needed API changes in another PR. 

I know there's more work to be done as we move the code base up to modern Rust standards, but I think it's probably better done across several than one huge PR.

secretfader:
Thanks @colinbankier. Glad to help. :+1:

PR #305, opened 2019-03-10
Migrate to Rust 2018 edition

whitfin:
This PR updates gotham/gotham_derive to Rust 2018. I didn't update the examples (yet), because I'm not sure if that might be "too soon" for general audiences.

### DIFF
--- a/examples/cookies/introduction/src/main.rs
+++ b/examples/cookies/introduction/src/main.rs
@@ -6,7 +6,7 @@ extern crate gotham;
 extern crate hyper;
 extern crate mime;
 
-use hyper::header::{COOKIE, SET_COOKIE};
+use hyper::header::SET_COOKIE;
 use hyper::{Body, Response, StatusCode};
 
 use cookie::{Cookie, CookieJar};
--- a/examples/cookies/introduction/src/main.rs
+++ b/examples/cookies/introduction/src/main.rs
@@ -66,7 +66,7 @@ pub fn main() {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use super::*;use hyper::header::COOKIE;
     use cookie::Cookie;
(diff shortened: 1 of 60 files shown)


colinbankier:
Thanks @whitfin . Looks like ci has failed - I'll let you check that out, looks like some missing `use crate::...` changes.

What's nice about this 2018 format this is that it makes it much more obvious what is from your own crate - I noticed a few cases we don't have these grouped together even though I think we tried to follow that pattern. Did you want to adjust the order/grouping of the `use` statements to tidy this up in this PR?

secretfader:
I've had some issues with removing `extern crate ...` declarations specifically for crates that use [proc-macro-hack](https://github.com/dtolnay/proc-macro-hack). When one macro imports another, the user is required to import *all of the macro's dependencies* as local, which makes the whole process quite messy.

tldr: So long as the macro import is clean, we should be clear to update examples and teach the new API going forward.

If not, my vote is to allow `extern crate` declarations for `gotham_derive` until we find a better way forward, because even the official docs on path clarity explain it will only work for roughly [99% of use cases.](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/path-clarity.html)

whitfin:
@colinbankier yep, I'll take a pass on the groupings :)

@secretfader I think it should be fine, I'll check it out

secretfader:
Excellent, @whitfin. Looks like you may also need to `use serde_derive::Deserialize` in a few places to satisfy CI. (I mention this because I've ran into similar errors and spent more time debugging than I should've.)

whitfin:
@secretfader yeah, I just missed a couple of imports in some test modules :).
@colinbankier I'm not going to update those groupings just yet; I'd rather do that post-merge I think.

PR #308, opened 2019-03-23
Serve tls

nyarly:
Still to do: TLS test module.


gotham/src/lib.rs
```diff
@@ -34,111 +34,56 @@ pub mod pipeline;
...
-use service::GothamService;
 
 /// Starts a Gotham application with the default number of threads.
-pub fn start<NH, A>(addr: A, new_handler: NH)
-where
-    NH: NewHandler + 'static,
-    A: ToSocketAddrs + 'static,
+/// If `tls_config` is `None`, the resulting service will run on HTTP,
+/// otherwise on HTTPS, which is preferred.
+pub fn start<NH, A>(addr: A, new_handler: NH, tls_config: Option<rustls::ServerConfig>)
```
whitfin: I know this PR is still WIP, but I'm not sure on this. I think it would be better to have `start` and `start_tls` rather than using the `Option.

nyarly: I opened the draft PR to solicit feedback before I was done, so that's more that okay.

What's your reasoning behind `start_tls`? First, note that the existing `start` became `plain::start` and there's an equivalent `tls::start`. 

My thinking is that a) we do want secure by default, (security being the 3rd focus of the project), so I wanted folks to at least notice that TLS is available and encourage them to switch to it. Being able to add a (shameful?) `None` to their existing code might be enough of a stumble to get attention.

At first, the intention here was to start in a "hybrid" mode, but it was quickly apparent that that would change the interface here significantly, since you need two different `SocketAddr`s for that. With that, and the TLS config, in mind, I start to wonder about a `gotham::ServerConfig` with a builder etc.

whitfin: Well yeah, either both, or have a default.

I personally thing the default should be what it currently is, just because it's the most familiar to people. If I'm trying to use Gotham and the docs point me to an API that requires me to set up certs and what not, it might be a turn off. Personally I typically use TLS in my proxy layer anyway (and I think that's quite a common pattern?).

A Gotham config isn't a bad idea as it masks a lot of it, probably better in the long run anyway :)

colinbankier: I too would commonly terminate TLS at a load balancer or proxy, and I think keeping the current default makes sense. As long as TLS is nice and easy to enable, I don't see this as compromising the goal of security as a core focus. Being pushed to deal with it for local development would be a pain.

nyarly: Switched this over to using plain::start

colinbankier:
Great work @nyarly. I had a read through, generally looks great, nothing jumped out at me to comment on, although I'm not super familiar with all the pieces that have changed. I'll wait for @whitfin to see if he has any more constructive feedback.

colinbankier:
@whitfin - any comments you wanted to make on this before merging?
@nyarly - Looks like this needs a quick rebase due to conflicts.

whitfin:
This looks good to me, the only thing I would maybe change is `pub use plain::*` to export the existing API as is for now - with TLS then being available under `gotham::tls`.

nyarly:
I'm going to merge this and follow up shortly with a PR for that change in interface.

PR #310, opened 2019-04-03
Support graceful shutdown

nyarly:
This is almost entirely @im-0's work. I've fixed up the example so that it works with new tokio/gotham, and it would be ready to merge as is.

The demonstrated feature of Gotham is that a web service can be interrupted and shut down gracefully. In the example program, this is done with an interrupt signal (as if the user had hit Ctrl-C), but the general pattern of `select`ing the server Future with another Future so that there's an outside termination behavior is pretty general. One could use a `oneshot` or a special HTTP endpoint to do the same thing.


Cargo.toml
```diff
@@ -8,10 +8,10 @@ members = [
     ## Middleware
     "middleware/template",
     "middleware/under_development/diesel",
-    "middleware/jwt",
```
colinbankier: I know we had to temporarily comment this out on #308 - was there a reason to remove it here as well? I liked leaving it as commented as reminder to bring it back when dependencies can be sorted out.

PR #313, opened 2019-04-30
Diesel tokio blocking middleware

### LINKED ISSUE #309: Async Diesel middleware and example
The current "under development" Diesel middleware provides a "blocking" api to use Diesel.
Some previous work was done (and abandoned) to provide a worker `CpuPool` middleware to allow not blocking request handling when using diesel.

I have been experimenting with continuing this, but with a different approach. Although Diesel doesn't directly support async, we can still build an async application around it using `tokio_threadpool::blocking`. This removes the need to explicitly manage a separate `CpuPool`, and looks like it will be a simpler implementation and api.

I'm opening this issue to solicit discussion around this approach and coordinate efforts in pushing Diesel support forward.
The new experimental api so far provides a `repo` abstraction that handles the DB connection pool and wraps `tokio_threadpool::blocking`. Usage looks something like:

```rust    
    let repo = Repo::borrow_from(&state);
    let user = .....;
    // Returns a future able to be chained into a `HandlerFuture`
    repo.run(move |conn| {
        diesel::insert_into(users::table)
            .values(&user)
            .get_result(&conn)
    }) 
```

Previous work this would supersede includes:
https://github.com/gotham-rs/gotham/pull/227
https://github.com/gotham-rs/gotham/issues/77
https://github.com/gotham-rs/gotham/pull/126

colinbankier:
Diesel middleware to close #309


colinbankier:
This pretty much includes all the things I initially intended.
@whitfin @nyarly - any thoughts or concerns before considering merging this in?

One thing I'm not sure of is how to handle specifying versions and dependency versions of middleware.
In order to create an app using a pre-release version of gotham, the middleware versions and their dependencies on gotham all need to line up. When we do a release, I guess these just all need to be updated for the release, and then for the next pre-release version. I'm not sure if this is captured in the current process.

colinbankier:
Thanks @nyarly . I had initially included this intentionally  - to demonstrate running tests with test transactions, and there was no good place to include the "create database" step before running tests.

I'm not sure this added much value - so have shifted these tests to use an in-memory db and deleted the tests.db. The example still shows how to use test transactions - it's just that they're redundant when the db is in-memory. I think that's ok - but could switch back if anyone prefers.

middleware/diesel/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 name = "gotham_middleware_diesel"
-version = "0.1.0"
-authors = ["Shaun Mangelsdorf <s.mangelsdorf@gmail.com>",
-           "Bradley Beddoes <bradleybeddoes@gmail.com>"]
+version = "0.4.0-dev"
```
whitfin: I think you mentioned this, but I don't know if we need to keep the versions in sync with Gotham for the sake of it... I'm leaning to `0.1.0` making more sense as a start point here - not sure if you agree?

middleware/diesel/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    T: Connection + 'static,
+{
+    fn clone(&self) -> Self {
+        match catch_unwind(|| self.repo.clone()) {
+            Ok(repo) => DieselMiddleware {
+                repo: AssertUnwindSafe(repo),
+            },
+            Err(_) => {
+                error!("PANIC: r2d2::Pool::clone caused a panic");
+                eprintln!("PANIC: r2d2::Pool::clone caused a panic");
```
whitfin: Think you might have left this `eprintln!` accidentally?

middleware/diesel/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn new_middleware(&self) -> io::Result<Self::Instance> {
+        match catch_unwind(|| self.repo.clone()) {
+            Ok(repo) => Ok(DieselMiddleware {
+                repo: AssertUnwindSafe(repo),
+            }),
+            Err(_) => {
+                error!(
+                    "PANIC: r2d2::Pool::clone caused a panic, unable to rescue with a HTTP error"
+                );
+                eprintln!(
```
whitfin: Same goes for this one

middleware/diesel/src/repo.rs
```diff
@@ -0,0 +1,172 @@
...
+        // `f.take()` allows the borrow checker to be sure `f` is not moved into the inner closure
+        // multiple times if `poll_fn` is called multple times.
+        let mut f = Some(f);
+        poll_fn(move || blocking(|| (f.take().unwrap())(pool.get().unwrap()))).then(
+            |future_result| match future_result {
+                Ok(query_result) => match query_result {
+                    Ok(result) => future::ok(result),
+                    Err(error) => future::err(error),
+                },
+                Err(e) => panic!(format!("Error running async database task: {:?}", e)),
```
whitfin: You shouldn't need the `format!` in here; `panic!` will format for you.

middleware/diesel/src/repo.rs
```diff
@@ -0,0 +1,172 @@
...
+
+impl<C, E> CustomizeConnection<C, E> for TestConnectionCustomizer
+where
+    C: diesel::connection::Connection,
+    E: std::error::Error + Sync + Send,
+{
+    fn on_acquire(&self, conn: &mut C) -> Result<(), E> {
+        match conn.begin_test_transaction() {
+            Ok(_) => Ok(()),
+            Err(_) => Ok(()),
```
whitfin: Does this maybe need a log or something? If it's only going to return `Ok(())`, then you can skip the match:

```
conn.begin_test_transaction();
Ok(())
```

middleware/jwt/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gotham_middleware_jwt"
-version = "0.3.1"
+version = "0.4.0-dev"
```
whitfin: Same comment on this as the other version; not sure if we need to align them.

middleware/diesel/Cargo.toml
colinbankier: The issue that I encountered was that the version of gotham declared by the middleware must match the version of gotham declared by your project (or cargo has be able to resolve them to the same version at least). This leaves the user with the awkward question of "which version of the middleware must I use?" - e.g. middleware X version "0.1" is compatible with my gotham version of "0.3", and middleware Y version "0.5" is. When gotham gets a new release, a new version of middlewares needs to be released that bumps the dependency version (with our pre-1.0 versions anyway). It all seemed easier to keep track of if the version numbers were just in sync - at least the "major" version.

middleware/diesel/src/lib.rs
colinbankier: Indeed, removed.

middleware/diesel/src/lib.rs
colinbankier: Removed.

middleware/diesel/src/repo.rs
colinbankier: fixed

middleware/diesel/src/repo.rs
colinbankier: Yep - good idea to log something meaningful at least. updated.

middleware/diesel/Cargo.toml
colinbankier: I've set this version back to "0.1.0".

middleware/jwt/Cargo.toml
colinbankier: I've set this back - but bumped the patch version as the gotham dependency version has changed.

colinbankier:
@whitfin  @nyarly - I think all the points above have been addressed.
Anything left here?

whitfin:
@colinbankier awesome! I'm going to merge and we can start playing with it!

PR #314, opened 2019-05-01
Re-add jwt middleware.

### LINKED ISSUE #311: Re-add JWT middleware
The JWT middleware depends on a crate that fixes it's version of Ring below what tokio_rustls needs. When #308 lands, we'll need to figure out how to get JWT back.

Options are:
* Wait, since presumably this'll be fixed upstream
* Patch the upstream crate - probably it'll just work
* Add a "tls" feature that enables TLS but disables JWT

I like patching, just because it gets the feature out the door, and we can remove the patch later.

colinbankier:
fixes #311 
A new version of jsonwebtoken has been published, and the conflict is resolved.


nyarly:
If you wait long enough by the banks of a river, the bodies of all your enemies will drift by.

PR #329, opened 2019-07-17
Fix a broken link in the examples README

idursun:
Link to `shared_state` example was mistyped and was returning 404.


whitfin:
@idursun awesome, thanks :D

PR #331, opened 2019-07-19
Fix an issue with missing cookie values

### LINKED ISSUE #330: Unable to receive multiple cookies
Using gotham `0.4.0` on crates.io

When doing the [cookie example](https://github.com/gotham-rs/gotham/tree/master/examples/cookies/introduction), the `CookieParser` seems unable to separate out multiple cookies?

E.g.
```
let temp = CookieJar::borrow_from(&state);
dbg!(&temp);
```

when given multiple cookies will output 

```
[src\main.rs:19] &temp = CookieJar {
    original_cookies: {
        DeltaCookie {
            cookie: Cookie {
                cookie_string: Some(
                    "othercookie=foobar; adjective=repeat;",
                ),
                name: Indexed(
                    0,
                    6,
                ),
                value: Indexed(
                    7,
                    13,
                ),
                expires: None,
                max_age: None,
                domain: None,
                path: None,
                secure: None,
                http_only: None,
                same_site: None,
            },
            removed: false,
        },
    },
    delta_cookies: {},
}
```

it seems like cookie parser doesn't separate out the cookies, leading it to only recognize the first cookie?

I just started with gotham, so I might be missing something here.

whitfin:
This fixes #330.

There's a dumb bug where it's only parsing the first cookie per header; this ensures that it parses all cookies. Very simple change, slightly annoying it got through the first time :p

### DIFF
--- a/gotham/src/middleware/cookie/mod.rs
+++ b/gotham/src/middleware/cookie/mod.rs
@@ -24,6 +24,7 @@ impl CookieParser {
             .get_all(COOKIE)
             .iter()
             .flat_map(HeaderValue::to_str)
+            .flat_map(|cs| cs.split("; "))
             .flat_map(|cs| Cookie::parse(cs.to_owned()))
             .fold(CookieJar::new(), |mut jar, cookie| {
                 jar.add_original(cookie);


gotham/src/middleware/cookie/mod.rs
```diff
@@ -24,6 +24,7 @@ impl CookieParser {
             .get_all(COOKIE)
             .iter()
             .flat_map(HeaderValue::to_str)
+            .flat_map(|cs| cs.split("; "))
```
colinbankier: It would seem safer to split on ";" and then `trim`. I guess most browsers insert a space, but I'm not sure it's technically required.

colinbankier:
Opened https://github.com/gotham-rs/gotham/pull/332 - if we get this merged and then rebase CI should be fixed for this PR.

whitfin:
@colinbankier I based `" ` on [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie), which seems like it is saying the standard has a space.

PR #338, opened 2019-08-02
Match documentation status codes with implementation

orva:


pksunkara:
@colinbankier This should be merged.

Issue #307, opened 2019-03-20
hyper Http structs ought to be bound to our executor


nyarly:
https://docs.rs/hyper/0.12.25/hyper/server/conn/struct.Http.html#method.with_executor

Seems that hyper uses the background tokio executor by default, which would break the contract implied by start_on_executor.

PR #342, opened 2019-08-12
Error: Blocking client in future context

ernestas-poskus:
```
examples/hello_world/
```

```
$ rustc -V
rustc 1.38.0-nightly (07e0c3651 2019-07-16)
```

```
$ cargo run
   Compiling gotham_examples_hello_world v0.0.0 (/rust/gotham/examples/hello_world)
    Finished dev [unoptimized + debuginfo] target(s) in 3.98s
     Running `/rust/gotham/target/debug/gotham_examples_hello_world`
Listening for requests at http://127.0.0.1:7878
Error(
    BlockingClientInFutureContext,
    "https://httpbin.org/ip",
)
Error(
    BlockingClientInFutureContext,
    "https://httpbin.org/ip",
)
```

Getting rather odd error when doing foreign 'sync' requests using reqwest crate in any handler.

Maybe this issue is related to https://github.com/gotham-rs/gotham/issues/307?

Tried to use a custom executor for Gotham server but got the same result. 

Maybe someone has had a similar error? Can't find much docs about it or true origin to debug further.

### DIFF
--- a/examples/hello_world/Cargo.toml
+++ b/examples/hello_world/Cargo.toml
@@ -9,3 +9,4 @@ gotham = { path = "../../gotham" }
 
 hyper = "0.12"
 mime = "0.3"
+reqwest = "0.9.19"
--- a/examples/hello_world/src/main.rs
+++ b/examples/hello_world/src/main.rs
@@ -3,6 +3,7 @@
 extern crate gotham;
 extern crate hyper;
 extern crate mime;
+extern crate reqwest;
 
 use gotham::state::State;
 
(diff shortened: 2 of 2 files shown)


colinbankier:
To use `reqwest` in a future context, use their async client.
Here is their example from https://github.com/seanmonstar/reqwest/blob/master/examples/async.rs translated into a gotham handler:

```
pub fn say_hello(state: State) -> Box<HandlerFuture> {
    let f = Client::new()
        .get("https://hyper.rs")
        .send()
        .and_then(|mut res| {
            println!("{}", res.status());

            let body = mem::replace(res.body_mut(), Decoder::empty());
            body.concat2()
        })
        .map_err(|err| println!("request error: {}", err))
        .map(|body| {
            let mut body = Cursor::new(body);
            let _ = io::copy(&mut body, &mut io::stdout()).map_err(|err| {
                println!("stdout error: {}", err);
            });
        });

    Box::new(f.then(move |_result| {
        let res = create_response(&state, StatusCode::OK, mime::TEXT_PLAIN, HELLO_WORLD);
        Ok((state, res))
    }))
}
```

ernestas-poskus:
thank you

PR #347, opened 2019-08-15
Fix the problem with guess_mime_type  deprecated and Add multipart examples

paulzhang5511:


examples/handlers/multipart/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+             ",
+            boundary
+        );
+
+        let test_server = TestServer::new(router()).unwrap();
+        let response = test_server.client()
+            .post("http://localhost", body, mime::MULTIPART_FORM_DATA)
+            .perform()
+            .unwrap();
+        // let header_map = request.headers_mut();
```
colinbankier: Here is how you add a header to your request:

```
        let test_server = TestServer::new(router()).unwrap();
        let client = test_server.client();
        let mut request = client.post("http://localhost", body, mime::MULTIPART_FORM_DATA);

        let headers = request.headers_mut();
        let h = format!("multipart/form-data; boundary={}", boundary);
        headers.insert(CONTENT_TYPE, HeaderValue::from_str(h.as_str()).unwrap());
        let response = request.perform().unwrap();
```

examples/handlers/multipart/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use gotham::test::TestServer;
+
+    #[test]
+    fn form_request() {
+        let boundary = "--abcdef1234--";
+        let body = format!(
+            "\
```
colinbankier: When I tested this out, it didn't like this extra new line. This worked for me:

```
        let body = format!(
            "--{0}\r\n\
             content-disposition: form-data; name=\"foo\"\r\n\r\n\
             bar\r\n\
             --{0}--\r\n",
            boundary
        );
```

examples/handlers/multipart/src/main.rs
paulzhang5511: @colinbankier  Thanks.  fixed.

paulzhang5511: this example to be in the same PR as the guess_mime_type fix

examples/handlers/multipart/src/main.rs
paulzhang5511: I use `cargo fmt`. but you can modify it.

examples/handlers/multipart/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "gotham_examples_handlers_multipart"
+description = "An example of decoding requests from an HTML form element"
+version = "0.0.0"
+authors = ["Jacob Budin <self@jacobbudin.com>"]
```
colinbankier: Looks like a few of these fields missed being updated when copied from the url encoded example - feel free to add your own name as the author, and let's update the description to specifically say "multipart" to differentiate it from the url encoded form example.

examples/handlers/multipart/src/main.rs
```diff
@@ -0,0 +1,126 @@
+//! An example of decoding requests from an HTML form element
```
colinbankier: Let's also specify "multipart" in here to make it clear how it's different from the other form example.

examples/handlers/multipart/Cargo.toml
paulzhang5511: @colinbankier You can modify it.

examples/handlers/multipart/src/main.rs
paulzhang5511: @colinbankier You can modify it.

examples/handlers/multipart/Cargo.toml
colinbankier:
```suggestion
authors = ["Paul Zhang <paulzhang5511>"]
```

examples/handlers/multipart/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "gotham_examples_handlers_multipart"
+description = "An example of decoding requests from an HTML form element"
```
colinbankier:
```suggestion
description = "An example of decoding multipart form requests"
```

examples/handlers/multipart/Cargo.toml
colinbankier: @paulzhang - it's easier if you update the last few things - otherwise I have to clone your repo and open a separate PR from my own branch. Committing Github PR suggestions directly doesn't work when I don't have write access to your repo.

colinbankier: I added some suggestions - if you could just select "commit suggestion" that'd be great :)

PR #349, opened 2019-08-25
Fix #348: Support trailing slashes when building a route

### LINKED ISSUE #348: Trailing slashes silently fails when adding a route
When adding a route with a trailing slash, gotham "silently fails". There's no failure per say, but the route is unreachable.

Let's consider the following minimal example:

```rust 
use gotham::state::State;
use gotham::router::Router;
use gotham::router::builder::{build_simple_router, DrawRoutes, DefineSingleRoute};


fn handler(state: State) -> (State, String) {
    (state, "Found!!!".into())
}

fn router() -> Router {
    build_simple_router(|route| {
        route.get("/foo/").to(handler);
    })
}

fn main() {
    let addr = "127.0.0.1:8080";
    println!("Listening for requests on http://{}/", addr);
    gotham::start(addr, router())
}
```

Both these curl command fail:

```
antoine@localhost$ curl -v  localhost:8080/foo
> GET /foo HTTP/1.1
< HTTP/1.1 404 Not Found
antoine@localhost$ curl -v  localhost:8080/foo/
> GET /foo/ HTTP/1.1
< HTTP/1.1 404 Not Found
```

After investigation, it looks like the `RouterBuilder` just split [the given path on `/`.](https://github.com/gotham-rs/gotham/blob/master/gotham/src/router/builder/draw.rs#L878)

If I add the paths `/foo/bar/` and `/foo/baz/`, this creates the following matching tree:

```
        +-----+
        |"foo"|
        +--+--+
           |
           |
 +-----+   |    +-----+
 |"baz"+<--+--->+"bar"|
 ++----+        +----++
  |                  |
+-v+                +v-+
|""|                |""|
+--+                +--+
```

However when walking the tree, the request uses `RequestPathSegments` which
[eliminates the empty segments.](https://github.com/gotham-rs/gotham/blob/2837acca37571fc76e65fa83e678c6858e64f1a4/gotham/src/helpers/http/request/path.rs) Therefore `GET /foo/baz/" is translated to `GET /foo/baz`. It never walks the tree to the empty string.

I see to solutions:

  * Failing at runtime (not optimal)
  * Uniformizing how the tree is built and walked
  * Create a way to generate paths that are safe, the same way serde_json does `serde_json::value!()`

acatton:
Uniformizes the parsing of path segments between `RouterBuilder` and
`RequestPathSegments`.

This fixes #348

### DIFF
--- a/gotham/src/helpers/http/request/path.rs
+++ b/gotham/src/helpers/http/request/path.rs
@@ -12,6 +12,10 @@ pub struct RequestPathSegments {
     segments: Vec<PercentDecoded>,
 }
 
+pub(crate) fn split_path_segments<'a>(path: &'a str) -> impl Iterator<Item = &'a str> {
+    path.split('/').filter(|s| !EXCLUDED_SEGMENTS.contains(s))
+}
+
 impl RequestPathSegments {
     /// Creates a new RequestPathSegments instance by splitting a `Request` URI path.
     ///
@@ -23,9 +27,7 @@ impl RequestPathSegments {
     /// ["/", "some", "path", "to", "my", "handler"]
     /// ```
     pub(crate) fn new(path: &str) -> Self {
-        let segments = path
-            .split('/')
-            .filter(|s| !EXCLUDED_SEGMENTS.contains(s))
+        let segments = split_path_segments(path)
             .filter_map(PercentDecoded::new)
             .collect();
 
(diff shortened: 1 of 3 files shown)


acatton:
Thanks for the quick merge @colinbankier ;) .

colinbankier:
Thanks for the contribution :+1:

PR #354, opened 2019-09-03
WIP Websocket support

vickenty:
This adds minimal support for websockets to gotham.

More precisely, it enables protocol upgrades on hyper requests, which is sufficient to handle websocket requests from the user code, without adding anything else to gotham.

While this make it possible to handle websockets (or other protocol upgrades), it doesn't make it ergonomic. Websocket implementations that I tried to use (rust-websocket and tungstenite) do not seem to provide a way to integrate their handshake implementations with Hyper (even when websockets uses hyper internally).

In the example I ended up implementing the handshake part myself, handing the upgraded connection to tungstenite to take care of the actual protocol. It works, but it is hardly an example to follow.

So I would like to use this PR to share my findings, and collect feedback about which direction to take this. Hopefully both rust-websocket and tungstenite can be improved to play along with hyper.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,9 @@ members = [
     "examples/diesel",
 
     # example_contribution_template
-    "examples/example_contribution_template/name"
+    "examples/example_contribution_template/name",
+
+    "examples/websocket",
 ]
 
 [patch.crates-io]
--- a/examples/websocket/Cargo.toml
+++ b/examples/websocket/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "websocket"
+version = "0.1.0"
+authors = ["Vickenty Fesunov <kent@setattr.net>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+gotham = { path = "../../gotham" }
+hyper = "0.12"
+futures = "0.1"
+tokio-tungstenite = "0.9"
+pretty_env_logger = "0.3"
+sha1 = "*"
+base64 = "*"
--- a/examples/websocket/README.md
+++ b/examples/websocket/README.md
@@ -1,0 +1,29 @@
+# Websocket
+
+Shows how to use Gotham with Tungstenite
+
+## Running
+
+From the `examples/websocket` directory:
+
+```
+Terminal 1:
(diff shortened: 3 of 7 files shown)


colinbankier:
Hi @vickenty, thanks for your initial investigations into this! I'd have to dig into this space a bit more, and take a look at what others are doing for ws support in order to give specific suggestions on where to go from here. Once we find a direction that we're happy with though, we can certainly pull some helper functions into gotham itself to ease the ergonomics for actual users.

Thanks for sharing your progress so far! This is a great start :)

svagner:
Hi! Is there any news on that? That patch is really required:)

colinbankier:
It's good to see there's strong interest in this @svagner :)
This is not quite ready for merging as is - but always keen to get input moving it forward.

It's not yet clear to me the trade-offs made between the major ws libraries (rust-websocket, ws-rs, tungstenite), and whether gotham should be coupled to any of them. Ideally the boilerplate stuff could be provided via an external crate, and gotham core just expose what is needed to make this possible.

@vickenty - are you familiar with whether the `with_upgrades()` call adds overhead or unwanted side-effects if the connection isn't to be upgraded? i.e. if this was called by default for all connections?

It will also be interesting to explore what support is required in the router to handle multiple ws endpoints along side normal routing.

colinbankier:
>  Ideally the boilerplate stuff could be provided via an external crate, and gotham core just expose what is needed to make this possible.
Although - I'm open to gotham picking one to use to provide built-in ws support if there is a strong case for one of them.

vickenty:
As far as I can tell, calling `with_upgrades()` by default shouldn't have much run-time impact (although only the profiler will tell for sure). Whether it is called or not, `hyper` does exactly the same thing, preparing for the eventual upgrade, except the future returned by `with_upgrades()` can take advantage of it (has an extra case in its `poll` impl).

It also seems to be possible to handle upgrades without using `with_upgrades()`, by polling the `Connection` future manually until it is exhausted, and then extracting the IO object. This might  be slightly less efficient, as this basically adds a layer around every connection object.

vickenty:
As for picking the implementation, it may be better to leave the question open for now (libraries are still evolving), and focus on making upgrades possible at all. Gotham can still help to reduce boilerplate by taking care of hyper side of things, like calling `body.on_upgrade()` and spawning new tasks.

PR #356, opened 2019-09-05
Expose generic `bind_server`

krallin:
This exposes a top-level bind_server and de-duplicates plain and tls
serving to use it. The underlying goal is to allow clients to use other
TLS implementations if they'd like.

There's a couple behavior changes this causes:

- This adds one extra function call when accepting plain connections,
  but that seems like it shouldn't amount to anything.
- This removes the panics in the TLS code path. Those could be triggered
  with an invalid handshake (e.g. a client connecting over HTTP to a
  HTTPS server), so it seems desirable to take them away.

This also updates Gotham to no longer _require_ rustls. It's still enabled by
default, but this lets users opt out of it. This lets us not depend on
rustls at all if we're not going to use it.


colinbankier:
Great, thanks for your contribution @krallin :)
Seems like worthwhile enhancements - since @nyarly provided the initial tls support, I'll let him give his thoughts.

Out of curiosity - did you have a use-case where you needed/wanted an alternative tls implementation?

krallin:
> Great, thanks for your contribution @krallin :)
> Seems like worthwhile enhancements - since @nyarly provided the initial tls support, I'll let him give his thoughts.

Thanks!

> Out of curiosity - did you have a use-case where you needed/wanted an alternative tls implementation?

Yes: we have an internal framework that provides a pre-configured `openssl::ssl::SslAcceptor` that works as per internal standards, so the goal was to be able to use that with Gotham :)

whitfin:
This seems good but I have a few questions/things I want to bring up:

1. I'm not sure whether `bind_server` should not have the third argument, and a second implementation be `bind_server_with_socket` which does. The former can call the latter. This eliminates `bind_server(X, Y,  |tcp| Ok(tcp).into_future())` being called by developers.
2. The trait bounds on the `W` function seem noisy. Could we maybe make the `W` type a `type` declaration in the top of the module with more extensive documentation? It seems like we need a good way to document this type if people are expected to provide it.
3. I think there are a few redundant `.map_err(|_| ())` and `.map(|_| ())` instances which should be removed.
4. You added `openssl` as a dependency, but I don't think it's used?

krallin:
> I'm not sure whether `bind_server` should not have the third argument, and a second implementation be `bind_server_with_socket` which does. The former can call the latter. This eliminates `bind_server(X, Y, |tcp| Ok(tcp).into_future())` being called by developers.

I'm guessing they'd probably want to call `init_server` in this case, wouldn't they? 

> The trait bounds on the `W` function seem noisy. Could we maybe make the `W` type a type declaration in the top of the module with more extensive documentation? It seems like we need a good way to document this type if people are expected to provide it.

Hmm; I'm not sure, this `W` is parameterized by `F`, which is what actually outlines what this function is supposed to return. I added a it more clarity on the `bind_server`  documentation, but if you have something specific in mind, I'd love to hear it!

> I think there are a few redundant `.map_err(|_| ())` and `.map(|_| ())` instances which should be removed.

I removed one that was indeed un-necessary. I think the other ones have to stay, since `executor::spawn` wants a future with no output. Do let me know if I missed one, though!

> You added `openssl` as a dependency, but I don't think it's used?

Thanks for pointing this out. I added it while testing but didn't mean to keep it in. I took it out.

> I'm not sure the rename to `bind_server_rustls` and making it public is necessary.

Making it public was indeed un-necessary (I modified this while trying out another way to make those changes but forgot to revert) — I fixed this now. Since it calls `bind_server, I renamed it so it wouldn't clash with `bind_server`

krallin:
I suspect the new test failures are due to https://github.com/rust-lang-nursery/futures-rs/pull/1744, which was released a couple days ago. I'll sent a separate PR to fix those.

krallin:
I updated this PR to also include an openssl example

Issue #358, opened 2019-09-25
Nightly CI breakage in gotham/src/middleware/session/mod.rs


o01eg:
After merging https://github.com/rust-lang/rust/pull/63565 in rustc, doctest in `gotham/src/middleware/session/mod.rs` had started to fail.

Soon it will became a beta and will break CI.

colinbankier:
Thanks for the report @o01eg

PR #360, opened 2019-10-07
Surface the source() fn in HandlerError

### LINKED ISSUE #359: HandlerError::into_response doesn't log cause
It looks like the code for `HandlerError::into_response` is supposed to log the cause of the error:

https://github.com/gotham-rs/gotham/blob/154034fbe92b5a7ff1b1732b6ddcdf4fb82b31ec/gotham/src/handler/error.rs#L140

However, that doesn't seem to work; on 0.4.0 my logs just show None where the cause description should be.

As far as I can tell it's because that line is calling `self.source()`, but `HandlerError` only implements `cause`, not `source`:

https://github.com/gotham-rs/gotham/blob/154034fbe92b5a7ff1b1732b6ddcdf4fb82b31ec/gotham/src/handler/error.rs#L79-L87

And checking `Error`'s default implementation, `source` will indeed return `None` by default.

Though, I haven't tested a patch to be sure that my theory is correct.

whitfin:
This might fix #359.

Basically we just surface the `source()` function from the `Error` trait correctly, by delegating to the internal error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: rust
 sudo: required
 dist: trusty
+addons:
+  apt:
+    packages:
+      - libssl-dev
 env:
   - PATH=$HOME/.cargo/bin:$PATH
 script:
@@ -14,9 +18,7 @@ matrix:
     # Run rustfmt in its own shard.
     - rust: stable
       env:
-        # Give a useful display name in the job list.
         - SHARD=rustfmt
-        - PATH=$HOME/.cargo/bin/:$PATH
       before_script:
         - rustup component add --toolchain stable rustfmt-preview
       script:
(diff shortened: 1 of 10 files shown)


colinbankier:
Thanks @whitfin. Seems fine - but travis seems unhappy.

whitfin:
@colinbankier the Travis unhappiness is due to https://github.com/gotham-rs/gotham/issues/358 and the fact it has moved into beta :p

fpgaminer:
I don't believe this fixes #359 

Here's my minimal test:

```
use gotham::handler::{HandlerError, IntoHandlerError};
use gotham::state::State;


fn main() {
	pretty_env_logger::init();
	gotham::start("127.0.0.1:7879", || Ok(say_hello))
}

fn say_hello(state: State) -> (State, Result<String, HandlerError>) {
	let err = std::io::Error::new(std::io::ErrorKind::Other, "oh no!");
	(state, Err(err.into_handler_error()))
}
```

To be run with: `RUST_LOG=debug cargo run`

Using this patch the log shows:

```
 DEBUG gotham::handler::error      > [8b96448c-5a73-4c05-9ece-e26f95c5aa41] HandlerError generating 500 Internal Server Error response: (none)
```

Still shows `(none)` where the error should be; the same behavior as 0.4.0.

I'm no expert on this code base or implementing `std:error::Error` but I don't believe this line is right:

```
self.cause.source()
```

`self.cause` is the "source" of the `HandlerError`.  So calling `self.cause.source()` actually returns the source of the source (which in my example usage is indeed None, hence the similar behavior).

My stab at it would be:
```
impl Error for HandlerError {
    fn description(&self) -> &str {
        "handler failed to process request"
    }

    fn cause(&self) -> Option<&dyn Error> {
        self.source()
    }

    fn source(&self) -> Option<&(dyn Error + 'static)> {
        Some(&*self.cause)
    }
}
```

Which appears to work correctly:

```
 DEBUG gotham::handler::error      > [4474b028-f506-4d12-a3ed-c8236f648d5d] HandlerError generating 500 Internal Server Error response: oh no!
```

Unless my assumptions are wrong, and the intention of `source` is to return the source of the source?

whitfin:
@fpgaminer you might be right, but then I'm a little confused what the point is to having both of those exist in the trait. What if `self.cause` has a `source` that is not `None`? Should we ignore it?

fpgaminer:
My understanding is that `source()` and `cause()` are supposed to be synonymous; exactly the same method.  `source()` only exists because the std devs had to patch a bug (users couldn't downcast the Error returned by `cause()` because it didn't specify a `'static` lifetime).  Beyond that, the intention of the method didn't change.  The naming of source versus cause is not significant.

So as a user I would expect for all implementations of `Error`, `some_error.source() == some_error.cause()`.

nyarly:
Nudge, @fpgaminer and @whitfin - what's the status here?

msrd0:
I think @fpgaminer is right - `Error::cause()` is marked deprecated in favour of `Error::source()`, so if we implement both, they should return the same result. Also I don't understand why this PR is making changes to the travis config.

msrd0:
I believe this has been fixed in #438.

PR #362, opened 2019-10-13
diesel middleware features: r2d2 instead of extras

msrd0:
There is no need to require features that aren't used by the middleware.


colinbankier:
Thanks for the PR @msrd0. You're right - but to TBH I'm wasn't 100% sure what cargo does if the middleware doesn't include some features, but then the user application does. I assume this would just work fine, but haven't tried it out. I'll see if I can do that just to be sure.

msrd0:
Well I'm pretty sure that cargo just merges the used features, at least I can tell that there are no problems having gotham not use the postgres-feature from diesel and my binary using it.

PR #363, opened 2019-10-17
Update url/percent-encoding to 2.1

msrd0:
Update the `url` dependency to 2.1 and replace with `percent-encoding` where necessary.

Dependabot tried but failed: https://github.com/gotham-rs/gotham/pull/340


colinbankier:
Excellent, thanks @msrd0 . Travis failed - it looks like just a minor rustfmt change.
After that, we're good to go :+1:

msrd0:
Sorry, for some reason GitHub just doesn't like to notify me when the CI fails. Should be fixed now.

Issue #355, opened 2019-09-05
Move Askama/Gotham integration into Gotham


djc:
I have been struggling with the Askama integrations for different web frameworks. For now, Askama is not available from crates.io for actix-web 1.0 because there are conflicting dependencies between actix-web/Gotham/Rocket, all of which Askama supports. Due to coherency issues, I think the integration code will have to live inside either Gotham or Askama. I'd still be happy to do any maintenance work required on the integration, but at this point I think it would be better if the Gotham/Askama integration moves into the Gotham crate (since the latter has many more requirements on the dependency tree than Askama itself).

See #320 of the previous instance of this.

msrd0:
It looks like the gotham-askama-integration is living inside a `askama_gotham` crate, which sounds like a good solution to me, so do you still want to move the integration code into gotham?

djc:
Yeah, I think the solution I came up with is working fairly well, so this can be closed.

PR #366, opened 2019-10-22
Add optional helper for creating responses from Askama templates

rakenodiax:
This PR adds `askama` as an optional dependency, which provides a helper function, `gotham::helpers::http::response::try_create_html_response`. This function creates a response with `StatusCode::OK` and `mime::TEXT_HTML_UTF_8`. The helper function returns a `Result`, because `askama` can return an error if `render` fails.

This is to address #355, but I wasn't able to get the same clean implementation of `IntoResponse` for `askama::Template` as the other implementation.


colinbankier:
Thanks @rakenodiax for the contribution.
I'm happy to include this in Gotham - particularly given that its behind a feature flag.
It'd be nice to update the example at https://github.com/gotham-rs/gotham/tree/master/examples/templating/askama to reflect this new integration.

@nyarly @whitfin any objections to having this in Gotham given #355 ?

whitfin:
I saw this and have been thinking on it. I'm not really sure I understand why it warrants inclusion; it seems quite trivial?

I'm hesitant to avoid it looking as though Gotham has "adopted" Askama as *the* templating engine.

colinbankier:
I had wondered the same thing - and we do already have an example. Maybe that is sufficient - this does seem like a particularly trivial function.
On the other hand, I'm ok with Gotham including explicit support for specific libraries to make getting started as easy as possible (whether templating, websockets, whatever) as long a user can opt-out and use something else if they prefer.

rakenodiax:
FWIW I think there are good arguments either way. I'd like to be able to implement `IntoResponse` directly on implementors of `Template`, but struggled with some orphan rule issues and it seems like coming at it from this direction is going to be difficult. As it is currently, Askama implements `IntoResponse` as a part of its codegen which avoids this issue.

As a user of Gotham, I like to be able to reduce boilerplate. I originally embarked on this to get the above capability, but decided that a helper function was more feasible. Either solution doesn't preclude additional template engine support, as well.

It seems like there's precedence to *not* include these types of functions in the crate, given that e.g. JSON isn't supported in this manner, either. If that's the direction y'all want to go, I completely understand, as well. :)

pksunkara:
```rust
pub fn render<D: fmt::Display>(state: State, view: D) -> (State, Response<Body>) {
    let mut content = String::new();

    let response = match write(&mut content, format_args!("{}", view)) {
        Ok(()) => create_response(
            &state,
            StatusCode::OK,
            mime::TEXT_HTML_UTF_8,
            content.into_bytes(),
        ),
        Err(_) => create_empty_response(&state, StatusCode::INTERNAL_SERVER_ERROR),
    };

    (state, response)
}
```

That is the helper function I have in my templating library for rendering a template into gotham.

msrd0:
#355 seems to be no longer necessary

PR #370, opened 2019-11-08
Migrate Gotham to std Futures to support Async/Await

kevinastone:
With the anticipated release of async-await in stable rust this week, I took an effort to migrate Gotham to run on std futures using the pre-release versions of the dependencies (tokio, hyper, etc).

*NOTE*: This doesn't attempt to introduce `await` or `async fn` into the codebase (there was a single unavoidable case due to an `tokio::File::metadata` API change).  That is designed as a future activity.

This migration involved a few key efforts:

1. Convert from Futures 0.1 to Futures-preview 0.3 (mostly `Future<Item=>` to `Future<Output=Result<>>`).
2. Update dependencies to pre-release versions (`tokio-0.2` and `hyper-0.13`).  There's still other dependencies that are outstanding and blocking the full release.
3. Migrate Handler trait to a pinned box HandlerFuture.

This is a **work-in-progress** with a few blockers before this would be ready:

Gotham Dependencies:

- [x] Update Futures from `futures-preview` to `futures = 0.3` when the other dependencies (hyper, tokio, etc) update in concert.
- [x] Update Tokio to `0.2` from alpha pre-releases
- [x] Update Hyper to `0.13` from alpha pre-releases
- [x] Update Tower-Service to `0.3` from alpha pre-releases.  Hyper is migrating many of its traits to `tower-service::Service` and so is now a direct dependency.
- [x] Released version of `futures_rustls` which is currently a branch of `tokio-rustls` ported to Futures-preview
- [x] Released version of `futures-tokio-compat` or suggested `tokio::compat` library for bridging `futures::AsyncRead` and `tokio::AsyncRead`.  See https://github.com/tokio-rs/tokio/issues/1297
    and https://github.com/async-rs/async-std/issues/54

Middleware Dependencies:

- [x] Diesel - Requires updated release of `tokio-threadpool`
- [ ] JWT - Requires updated release of `jsonwebtoken` since it's dependency `ring` conflicts withs `futures-rustls`.

The following examples were not updated (and thus commented out in `Cargo.toml`) due to incompatible dependencies:

- [x] `hello_world_until` is dependent on a compatible version of `tokio-signal`
- [x] `diesel` is dependent on the Diesel middleware
- [ ] `openssl` is dependent on `tokio-openssl`
- [ ] `websocket` is dependent on `tokio-tungstenite`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,13 +7,15 @@ members = [
 
     ## Middleware
     "middleware/template",
-    "middleware/diesel",
-    "middleware/jwt",
+    # TODO: Re-enable middleware when their dependencies are updated
+    # "middleware/diesel",
+    # "middleware/jwt",
 
     ## Examples (these crates are not published)
     "examples/hello_world",
     "examples/hello_world_tls",
-    "examples/hello_world_until",
+    # TODO: Re-enable when the tokio-signal dependency is updated
+    # "examples/hello_world_until",
     "examples/shared_state",
 
     # Tera template
@@ -65,15 +67,18 @@ members = [
     "examples/static_assets",
 
     # diesel
-    "examples/diesel",
+    # TODO: Re-enable when the middleware is updated
+    # "examples/diesel",
 
     # openssl
-    "examples/openssl",
+    # TODO: Re-enable when this example is updated
+    # "examples/openssl",
 
     # example_contribution_template
     "examples/example_contribution_template/name",
 
-    "examples/websocket",
+    # TODO: Re-enable when tokio-tungstenite is updated
+    # "examples/websocket",
 ]
 
 [patch.crates-io]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,7 @@ members = [
     ## Examples (these crates are not published)
     "examples/hello_world",
     "examples/hello_world_tls",
-    # TODO: Re-enable when the tokio-signal dependency is updated
-    # "examples/hello_world_until",
+    "examples/hello_world_until",
     "examples/shared_state",
 
     # Tera template
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,8 @@ members = [
 
     ## Middleware
     "middleware/template",
+    "middleware/diesel",
     # TODO: Re-enable middleware when their dependencies are updated
-    # "middleware/diesel",
     # "middleware/jwt",
 
     ## Examples (these crates are not published)
@@ -66,8 +66,7 @@ members = [
     "examples/static_assets",
 
     # diesel
-    # TODO: Re-enable when the middleware is updated
-    # "examples/diesel",
+    "examples/diesel",
 
     # openssl
     # TODO: Re-enable when this example is updated
--- a/examples/diesel/Cargo.toml
+++ b/examples/diesel/Cargo.toml
@@ -11,8 +11,8 @@ publish = false
 gotham = { path = "../../gotham/"}
 gotham_derive = { path = "../../gotham_derive/" }
 gotham_middleware_diesel = { path = "../../middleware/diesel"}
-hyper = { version = "0.13.0-alpha.4", features = ["unstable-stream"] }
-futures-preview = "0.3.0-alpha.19"
+hyper = "0.13.1"
+futures = "0.3.1"
 mime = "0.3"
 log = "0.4"
 diesel = { version = "1.4", features = ["sqlite", "extras"] }
@@ -25,4 +25,4 @@ serde_derive = "1.0"
 
 [dev-dependencies]
 diesel_migrations = "1.4.0"
-tokio = "=0.2.0-alpha.6"
+tokio = { version = "0.2.6", features = ["full"] }
--- a/examples/diesel/Cargo.toml
+++ b/examples/diesel/Cargo.toml
@@ -11,8 +11,8 @@ publish = false
 gotham = { path = "../../gotham/"}
 gotham_derive = { path = "../../gotham_derive/" }
 gotham_middleware_diesel = { path = "../../middleware/diesel"}
-hyper = "0.12"
-futures = "0.1"
+hyper = { version = "0.13.0-alpha.4", features = ["unstable-stream"] }
+futures-preview = "0.3.0-alpha.19"
 mime = "0.3"
 log = "0.4"
 diesel = { version = "1.4", features = ["sqlite", "extras"] }
@@ -25,4 +25,4 @@ serde_derive = "1.0"
 
 [dev-dependencies]
 diesel_migrations = "1.4.0"
-tokio = "0.1"
+tokio = "=0.2.0-alpha.6"
--- a/examples/diesel/Cargo.toml
+++ b/examples/diesel/Cargo.toml
@@ -12,6 +12,7 @@ gotham = { path = "../../gotham/"}
 gotham_derive = { path = "../../gotham_derive/" }
 gotham_middleware_diesel = { path = "../../middleware/diesel"}
 hyper = "0.13.1"
+failure = "0.1"
 futures = "0.3.1"
 mime = "0.3"
 log = "0.4"
--- a/examples/diesel/src/main.rs
+++ b/examples/diesel/src/main.rs
@@ -9,7 +9,7 @@ extern crate diesel_migrations;
 
 use diesel::prelude::*;
 use diesel::sqlite::SqliteConnection;
-use futures::{future, Future, Stream};
+use futures::prelude::*;
 use gotham::handler::{HandlerError, HandlerFuture, IntoHandlerError};
 use gotham::helpers::http::response::create_response;
 use gotham::pipeline::{new_pipeline, single::single_pipeline};
@@ -43,9 +43,9 @@ struct RowsUpdated {
     rows: usize,
 }
 
-fn create_product_handler(mut state: State) -> Box<HandlerFuture> {
+fn create_product_handler(mut state: State) -> Pin<Box<HandlerFuture>> {
     let repo = Repo::borrow_from(&state).clone();
-    let f = extract_json::<NewProduct>(&mut state)
+    extract_json::<NewProduct>(&mut state)
         .and_then(move |product| {
             repo.run(move |conn| {
                 // Insert the `NewProduct` in the DB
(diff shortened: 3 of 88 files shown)


colinbankier:
Wow - thanks @kevinastone! This is a huge effort! This is definitely something that would have been inevitable soon but we hadn't managed to start yet. We'll see if we can divide and conquer the remaining items on your list :+1:

alsuren:
After applying this patch, will `async fn` be able to impl Handler directly, or will we need to write a normal function which contains an async block, and then box the resulting future for returning, like I have to do in #281 ?

kevinastone:
Without async traits, I don't believe we can implement `async fn` handlers directly.  I wrote a proc-macro in my own code (happy to upstream) to allow annotating your async fn to generate the boilerplate async-block/boxed future for you: https://github.com/kevinastone/httpbox/blob/master/gotham_async/src/lib.rs

Here's an example:
https://github.com/kevinastone/httpbox/blob/gotham-future/src/app/method/body.rs#L47-L53

alsuren:
I have made a PoC of how I think we could make `async fn ...` impl `Handler` without any macros. I've published it as a PR against this branch: https://github.com/kevinastone/gotham/pull/1 .

Unfortunately, it had to break support for the `(State, T)` convenience return type (due to E0119), so it needs a bit of an API design discussion if we want to go in that direction:

* If we want to support the convenience (State, T) return type, we might be able to do this with a new .to_simple() function in the router builder. We might even be able to add a .to_compat() function, which takes an old-style future.
* Alternatively, we could keep the existing .to() method, and Handler types using old-style futures, and make a new `.to_async()` function that accepts a new-style `async fn`.
  * This would make the upgrade a little bit less painful for our users. We can then deprecate `.to()` at our leisure.
* I haven't looked into how this would affect middleware.

pksunkara:
I would vote for `.to_async()`

pksunkara:
So, I started working on this branch to tick off items above the list, but I got stuck with the following:

```
error[E0277]: the trait bound `tokio_rustls::client::TlsStream<tokio::net::tcp::stream::TcpStream>: hyper::client::connect::Connection` is not satisfied
   --> gotham/src/tls/test.rs:170:5
    |
170 | /     pub fn client(&self) -> TestClient<Self, TestConnect> {
171 | |         self.client_with_address(SocketAddr::new(IpAddr::from([127, 0, 0, 1]), 10000))
172 | |     }
    | |_____^ the trait `hyper::client::connect::Connection` is not implemented for `tokio_rustls::client::TlsStream<tokio::net::tcp::stream::TcpStream>`
    | 
   ::: gotham/src/test.rs:83:1
    |
83  |   pub struct TestClient<TS: Server, C: Connect> {
    |   --------------------------------------------- required by `test::TestClient`
    |
    = note: required because of the requirements on the impl of `hyper::client::connect::sealed::Connect` for `tls::test::TestConnect`
```

I can't implement the above hyper trait for a tokio_rustls struct.

My branch is at: https://github.com/reign-rs/gotham/tree/await

kevinastone:
It was a considerable amount of work to port to `tokio=0.2` and `hyper=0.13`.  Here's a branch with my WIP changes:

https://github.com/gotham-rs/gotham/commit/d6c11a601931e1246beeaac5325f883f3774bf27

It seems to run correctly, I just haven't got all the doctests to run successfully.  I'd also like to split up the commits into something more manageable.

colinbankier:
Thanks for the efforts here so far @kevinastone @pksunkara .
Since this is a BIG change, with breaking api changes, I'd certainly want @whitfin and @nyarly to weigh in with their thoughts.

nyarly:
I'll queue up a review by EOW.

kevinastone:
I pushed up one big commit for the final `Tokio-0.2` and `hyper-0.13` migration to unblock others.  I haven't had the chance to do any desirable refactoring and investigate some of the test failures (or really, tests that never complete).

gotham/src/tls/test.rs
```diff
@@ -83,18 +88,31 @@ impl Clone for TestServer {
 
 impl test::Server for TestServer {
     fn request_expiry(&self) -> Delay {
-        Delay::new(Instant::now() + Duration::from_secs(self.data.timeout))
+        let runtime = self.data.runtime.write().unwrap();
+        runtime.enter(|| delay_for(Duration::from_secs(self.data.timeout)))
+        // delay_for(Duration::from_secs(self.data.timeout))
```
pksunkara:
```suggestion
```

kevinastone:
The JWT middleware is still blocked on an updated version of `ring`.  It doesn't seem to be maintained: https://github.com/Keats/jsonwebtoken#help-wanted-for-v7

Tokio-openssl (required by `"examples/openssl"`) has been parked and expected to be integrated into rust-openssl, but no timeline is available: https://github.com/alexcrichton/tokio-openssl/issues/12

Tokio-tungstenite (required by `"examples/websocket"`) has not been updated to tokio-0.2 yet: https://github.com/snapview/tokio-tungstenite/issues/64.  It's last commit is 6 months old, so it might be a while before a compatible update is available.

Besides these middleware/examples above, the rest of the outstanding items for this PR are complete.

kevinastone:
@ping @nyarly and @whitfin for review.

pksunkara:
I have reviewed it pretty extensively. I don't see any issues with the code.

Cargo.toml
```diff
@@ -8,7 +8,8 @@ members = [
     ## Middleware
     "middleware/template",
     "middleware/diesel",
-    "middleware/jwt",
+    # TODO: Re-enable middleware when their dependencies are updated
+    # "middleware/jwt",
```
whitfin: If this is going to be merged with these commented out, can we get issues for each one which needs updating? I know there are notes in this PR, but that's not as simple for a beginner to pick up, should they choose to.

examples/diesel/Cargo.toml
```diff
@@ -25,4 +26,4 @@ serde_derive = "1.0"
 
 [dev-dependencies]
 diesel_migrations = "1.4.0"
-tokio = "0.1"
+tokio = { version = "0.2.6", features = ["full"] }
```
whitfin: What specifically do we need from the `full` feature that is missing from the default?

examples/diesel/src/main.rs
```diff
@@ -43,46 +44,50 @@ struct RowsUpdated {
...
+            Ok(rows) => rows,
+            Err(e) => return Err((state, e.into_handler_error())),
+        };
+
+        let body =
+            serde_json::to_string(&RowsUpdated { rows }).expect("Failed to serialise to json");
+        let res = create_response(&state, StatusCode::CREATED, mime::APPLICATION_JSON, body);
+        Ok((state, res))
+    }
+    .boxed()
```
whitfin: I haven't checked recently, but wasn't this form basically deprecated in favour of `Box::new`?

examples/diesel/src/main.rs
```diff
@@ -43,46 +44,50 @@ struct RowsUpdated {
...
-        .and_then(move |product| {
-            repo.run(move |conn| {
-                // Insert the `NewProduct` in the DB
+    async move {
+        let product = match extract_json::<NewProduct>(&mut state).await {
+            Ok(product) => product,
+            Err(e) => return Err((state, e)),
+        };
+
+        let rows = match repo
```
whitfin: Can we move the `repo.run` out of the `match` and into a local var? It's a little hard to parse at a glance, considering this is an example.

examples/diesel/src/main.rs
whitfin: Rather than repeat, this comment applies to all `.boxed()`.

examples/openssl/src/main.rs
```diff
@@ -41,15 +41,15 @@ pub fn main() -> Result<(), Error> {
 
     let listener = TcpListener::bind(&addr)?;
 
+    let mut runtime = Runtime::new()?;
```
whitfin: I'm not worried about it, but I'm curious why this move happened?

gotham/src/handler/assets/mod.rs
```diff
@@ -7,12 +7,15 @@
 mod accepted_encoding;
 
 use crate::error::Result;
+use bytes::Bytes;
```
whitfin: You can remove this line and combine it with the one below.

gotham/src/handler/assets/mod.rs
```diff
@@ -194,59 +198,62 @@ impl Handler for DirHandler {
...
-        Ok(response) => Ok((state, response)),
-        Err(err) => {
-            let status = match err.kind() {
-                io::ErrorKind::NotFound => StatusCode::NOT_FOUND,
-                io::ErrorKind::PermissionDenied => StatusCode::FORBIDDEN,
-                _ => StatusCode::INTERNAL_SERVER_ERROR,
-            };
-            Err((state, err.into_handler_error().with_status(status)))
+    let response_future = File::open(path).and_then(|file| {
+        async move {
```
whitfin: For my sake, because I'm not completely up to date with the syntax... can this become `|file| async move {`?

gotham/src/service/trap.rs
```diff
@@ -91,78 +96,19 @@ fn finalize_panic_response() -> FutureResult<Response<Body>, CompatError> {
...
+            .status(StatusCode::INTERNAL_SERVER_ERROR)
+            .body(Body::default())
+            .unwrap()
+    });
 
     future::ok(response)
 }
 
-/// Wraps a future to ensure that a panic does not escape and terminate the event loop.
-enum UnwindSafeFuture<F>
```
whitfin: \o/

gotham/src/handler/assets/mod.rs
kevinastone: Unfortunately no.  Async closures [are not stable yet][0].  Oddly, you can pass a closure and an async block in lieu.

[0]: https://rust-lang.github.io/rfcs/2394-async_await.html#async--closures

examples/diesel/src/main.rs
kevinastone: [`.boxed()`][0] is part of the `FutureExt` trait extension.  It's just a helper for `Box::pin`.  In fact, that's the [entire implementation][1].

[0]: https://docs.rs/futures/0.3.1/futures/future/trait.FutureExt.html#method.boxed
[1]: https://github.com/rust-lang/futures-rs/blob/f6ee3bdf925af52b67c4258618dad7e011c8f648/futures-util/src/future/future/mod.rs#L463

examples/diesel/Cargo.toml
kevinastone: Tokio includes [nothing by default][0] 😞.  You have to specify what features you need explicitly.  I believe we use:
- blocking
- net
- dns
- rt-core
- time
- fs (maybe?)

But I haven't tried to be more granular than just full.

[0]: https://github.com/tokio-rs/tokio/blob/eb1a8e1792b2c4b296be47a0681421c90bbdbf7a/tokio/Cargo.toml#L28

Cargo.toml
kevinastone: Do you want me to create these issues?  I'm not sure of the future of these middleware given the state of the dependencies.

examples/openssl/src/main.rs
kevinastone: I think it's unnecessary.  During the refactor, I originally passed the runtime to `bind_server`, but later refactored it out when I made `bind_server` an async function instead.

examples/diesel/src/main.rs
kevinastone: Sure.

pksunkara:
Thanks for the merge. Now we need to get #281 done so we can properly use async/await.

PR #372, opened 2019-11-12
Re-enable protocol upgrades

### LINKED ISSUE #353: WebSocket
Hi all

Is it possible to use a WebSocket with gotham?

Cheers,
Stefan

SchrodingerZhu:
this will fix #353.

**status**: passed my local test

Shall we add some tests on websocket/protocol upgrades?

### DIFF
--- a/gotham/src/lib.rs
+++ b/gotham/src/lib.rs
@@ -115,6 +115,7 @@ where
                 .and_then(move |socket| {
                     accepted_protocol
                         .serve_connection(socket, service)
+                        .with_upgrades()
                         .map_err(|_| ())
                 })
                 .map(|_| ());


colinbankier:
Thanks @SchrodingerZhu . Yes - I think we should add a test to catch such a regression in the future. Are you happy to add this? If you're unsure, perhaps @vickenty could suggest the simplest way to test this.

SchrodingerZhu:
I am glad to write some tests, but I don't quite get the test model of Gotham. Any instruction to follow?

msrd0:
This change has been implemented in https://github.com/gotham-rs/gotham/commit/2b60b47cb01ead1124b9d0063046f454be831900

PR #384, opened 2019-12-30
Update syn and quote to 1.0

pksunkara:
Fixes #345 
Fixes #346


pksunkara:
@colinbankier Should be easy to review since the fix is just those dependencies need to be upgraded together.

PR #410, opened 2020-02-29
Give priority to static children that are delegated

pksunkara:
The following code is a pretty common case when people are trying to do different apps with mix and match pipelines.

```rust
let delegated_router = build_simple_router(|route| {
    route.get("/b/a").to(welcome::delegated);
});

let child_delegated_router = build_simple_router(|route| {
    route.get("/a").to(welcome::index);
});

let router = build_simple_router(|route| {
    route.delegate("/").to_router(delegated_router);

    route.delegate("/b").to_router(child_delegated_router);
});
```

When I go to `/b/a`, the `welcome::delegated` is being run. Static children types should be given priority if they are delegated too. (Note: non-delegated children won't have priority, so, it still works as before)

Router tree and `match_node` needs a bigger overhaul if we want to support other type of children that are being delegated too. But that's for later.

### DIFF
--- a/gotham/src/router/builder/mod.rs
+++ b/gotham/src/router/builder/mod.rs
@@ -663,11 +663,11 @@ mod tests {
     #[test]
     fn test_delegated_priority_after_delegated_children() {
         let delegated_router = build_simple_router(|route| {
-            route.get("/b").to(welcome::delegated);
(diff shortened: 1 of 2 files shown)


pksunkara:
I did some research and I released this PR if implemented is actually going to encourage bad user behaviour. So, I will be closing this.

PR #403, opened 2020-02-19
Upgrade jsonwebtoken to 7.0 and re-enable

colinbankier:
Fixes #390

PR #415, opened 2020-03-07
Re-Enable JWT Middleware

### LINKED ISSUE #390: Re-enable JWT middleware after async/await
The JWT middleware was disabled in #370 and is still blocked on an updated version of jsonwebtoken due to a conflicting dependency on ring. It doesn't seem to be maintained: https://github.com/Keats/jsonwebtoken#help-wanted-for-v7

kevinastone:
`jsonwebtoken` updated to `v0.7` which is compatible with the `ring`
depeendency in the current master trunk.

Fixes #390.

### DIFF
--- a/middleware/jwt/Cargo.toml
+++ b/middleware/jwt/Cargo.toml
@@ -20,5 +20,5 @@ gotham = { path = "../../gotham", version = "0.5.0-dev" }
 gotham_derive = { path = "../../gotham_derive", version = "0.5.0-dev" }
 serde = "1.0"
 serde_derive = "1.0"
-jsonwebtoken = "6.0"
+jsonwebtoken = "^7.0"
 log = "0.4"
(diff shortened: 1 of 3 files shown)


middleware/jwt/Cargo.toml
```diff
@@ -20,5 +20,5 @@ gotham = { path = "../../gotham", version = "0.5.0-dev" }
 gotham_derive = { path = "../../gotham_derive", version = "0.5.0-dev" }
 serde = "1.0"
 serde_derive = "1.0"
-jsonwebtoken = "6.0"
+jsonwebtoken = "^7.0"
```
pksunkara: I don't think you need `^`

pksunkara:
There's already #403 btw

kevinastone:
> There's already #403 btw

Oh sorry!  Let's get #403 merged.

Issue #376, opened 2019-11-22
Failed to build example(s)


vitali2y:
No way to build all or single example unless to do `cargo build`:
```
vit@evergreen ~/prjs $ git clone https://github.com/gotham-rs/gotham.git
~...~
vit@evergreen ~/prjs $ cd ./gotham
vit@evergreen ~/prjs/gotham $ cargo build --examples
    Updating crates.io index
    Finished dev [unoptimized + debuginfo] target(s) in 2.58s
vit@evergreen ~/prjs/gotham $ ls -la ./target/debug/examples/
total 8
drwxrwxr-x 2 vit vit 4096 Nov 22 02:11 .
drwxrwxr-x 7 vit vit 4096 Nov 22 02:11 ..
vit@evergreen ~/prjs/gotham $ cargo build --example websocket
error: no example target named `websocket`
vit@evergreen ~/prjs/gotham $
```

msrd0:
Sorry for the long delay on this issue. Does #423 fix the issue or do you still have trouble building (individual) examples?

msrd0:
I'll close this for now, feel free to reopen if you continue experiencing issues with example compilation.

vitali2y:
Thank you - sorry, did not have time to re-test it.

PR #423, opened 2020-04-23
Fix the websocket example (and reenable hyper connection upgrades as a prerequisite)

### LINKED ISSUE #392: Re-enable websocket example after async/await
Websocket example was disabled in #370 .
Tokio-tungstenite (required by "examples/websocket") has not been updated to tokio-0.2 yet: snapview/tokio-tungstenite#64. It's last commit is 6 months old, so it might be a while before a compatible update is available.

FSMaxB:
Fixes #392 #376 #353 by incorporating the change from #372 and updating the websocket example to futures 0.3 and tokio 0.2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ members = [
     "examples/example_contribution_template/name",
 
     # TODO: Re-enable when tokio-tungstenite is updated
-    # "examples/websocket",
+    "examples/websocket",
 
     # finalizer
     "examples/finalizers/",
--- a/examples/websocket/Cargo.toml
+++ b/examples/websocket/Cargo.toml
@@ -9,7 +9,8 @@ edition = "2018"
 [dependencies]
 gotham = { path = "../../gotham" }
 futures = "0.3.1"
-tokio-tungstenite = "0.9"
+tokio-tungstenite = "0.10"
+tokio = "0.2"
 pretty_env_logger = "0.3"
 sha1 = "*"
 base64 = "*"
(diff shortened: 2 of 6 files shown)


nyarly:
I'm inclined to merge this, despite the apparent decline in coverage, since the coverage fails in the examples only.

@FSMaxB Is there any way you could exercise the example code in tests, though?

FSMaxB:
@nyarly Not sure how to interpret the coverage report. I don't quite understand how my change reduced coverage in anything but the examples.

I'm not sure how I could actually test this because I don't think the `TestServer`'s `TestClient` supports switching from http to a raw socket, which would be required in this case. (only alternative would be actually opening a port while running the tests, but that is kind of flakey.

FSMaxB:
I just took a closer look at the `TestClient` and it is possible to access a `Response<Body>` with an `on_upgrade` method, so it might be possible to test this after all. Probably won't find the time this weekend though.

FSMaxB:
I did it, I succeeded in a first proof of concept of using `gotham::test::TestServer` for websocket communication. Although it is currently quite awkward to use the `TestClient` and `TestServer`, especially when it comes to tests that require asynchronous code and upgrading hyper connections.

But more on that at some point later in the coming week.

FSMaxB:
I just added tests verifying the entire websocket example. This also incidentally tests that upgrading hyper sockets works.

FSMaxB:
Also I do not understand in the slightest what this codecov report is supposed to tell me.

@nyarly No idea why the tests are not showing up in the coverage, but it is definitely tested in the latest changes.

Issue #381, opened 2019-12-01
ContentTypeHeaderRouteMatcher matches too strict


msrd0:
I'm using gotham's `ContentTypeHeaderRouteMatcher` to only accept POST requests with `application/json` as their Content-Type. One application decided to send `application/json; charset=utf-8`. I'd expect this to be fine, however, gotham decides to answer with a _415 Unsupported Media Type_. I believe that gotham should behave less strict and accept the Content-Type even if additional information like a charset is given.

msrd0:
I've implemented my own version of the content type and accept header matchers [here](https://gitlab.com/msrd0/gotham-restful/-/tree/master/gotham_restful/src/matcher), if you're interested I could create a PR and replace gotham's more basic matchers with those. They'd allow content types to specify additional information like the charset, and allow accept headers to contain the quality-weighted syntax as returned by browsers rather than just a single mime type (which is fine for most rest applications but triggers 415 responses when accessed through browsers).

nyarly:
A quick review of that repo leaves me with a very good impression. I'd be inclined to merge such a PR. Thanks!

PR #424, opened 2020-04-25
New implementation of Content-Type and Accept header route matchers

msrd0:
I have reimplemented the Content-Type and Accept header matchers to suit more advanced needs. This includes a fix for #381. Both now use a `HashMap` for efficient lookup and also

**The Accept matcher**
- now supports more than one type from the header
- supports the quality-weighted syntax, although qualities are ignored
- will not only match `*/*` as a wildcard, but also types with subtype-wildcards like `image/*`
- will now ignore parameters like `charset=utf-8` as they are not supposed to be part of the Accept header, and conflict with the list syntax as both use commas as separators

**The Content-Type matcher**
- now matches content types in respect of their parameters, so that `text/plain; charset=utf-8` is considered a match when `text/plain` was expected, however, `text/plain; charset=utf-8` is not a match if `text/plain; charset=us-ascii` was expected
- allows the user to optionally specify that requests without a Content-Type header are acceptable, which is at least useful during development when your curl commands get shorter

Also, I have added public re-exports in their respective super mod for `ContentTypeHeaderRouteMatcher` and `RouteNonMatch` to simplify user code.

### DIFF
--- a/gotham/Cargo.toml
+++ b/gotham/Cargo.toml
@@ -48,6 +48,7 @@ cookie = "0.13"
 http = "0.2"
 httpdate = "0.3"
 failure = "0.1"
+itertools = "0.9.0"
 tokio-rustls = { version = "0.12.1", optional = true }
 
 [dev-dependencies]
--- a/gotham/src/extractor/internal.rs
+++ b/gotham/src/extractor/internal.rs
@@ -114,12 +114,12 @@ macro_rules! single_value_type {
     ($trait_fn:ident, $visitor_fn:ident) => {
         fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
         where
-            V: Visitor<'de>
+            V: Visitor<'de>,
         {
             let v = parse_single_value(self.values)?;
             visitor.$visitor_fn(v)
         }
-    }
+    };
 }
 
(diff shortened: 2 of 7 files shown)


msrd0:
Well, as I understand the task of the accept header matcher is to see whether there is a match, but has no way to inform the handler as to what quality that match is. Even worse, when a handler has several types it can produce, there is currently no way the matcher can inform the handler which of the types actually produces that match. This is most certainly an issue but sounds like it can't (currently) be solved through a route matcher.

Also, I have kept your documentation on both matcher, although the code contained in them reads more as a test than as a usage example. Let me know if you'd like to see this changed.

msrd0:
I just caught a case I hadn't considered yet and pushed a fix for it.

pksunkara:
@msrd0 I have a small question. This is just for matching, right? So, the router looks for matches and if the route doesn't match because of this, it would check other routes and at the end returns 404. But not 406 or 415. Or am I mistaken?

msrd0:
@pksunkara Well, I guess the idea is for gotham to return the `NonMatch` status code if this route didn't match and no other matching route was found. However, I found this behaviour to be inconsistent. For example, when I tried to return a `NonMatch` with status code 404, I got a 405 from the router instead (https://github.com/gotham-rs/gotham/issues/434).

PR #425, opened 2020-04-27
Remove unnecessary restrictions from test::Server::run_future

FSMaxB:
The requirements on the `test::Server` traits `run_future` method are unnecessarily strict in my opinion.

This removes the following:
* `'static` lifetime bound
* `Result` as output type (any output can be allowed)
* `Send` on the future and output (this is debatable, there are conceivable implementations of `Server` that might want this, so I'm okay for this to be left in if need be)

Removing these bounds is especially useful when working with the `TestClient`.

```rust
let client = server.client();
let foo: &SomeType = bar();
let first = server.run_future(async {
   foo.bar().await
});
let blub = client.get(something).perform(); // <- This can't be run inside of Server::run_future because blocking is not allowed in there
let second = server.run_future(async {
   foo.yadeyade(blub).await
});
```

This won't work without my change because `foo` doesn't have a static lifetime.

And having to manually wrap and unwrap `Result`s all the time isn't very ergonomic either, so removing `Result` is also quite useful.

### DIFF
--- a/gotham/src/extractor/internal.rs
+++ b/gotham/src/extractor/internal.rs
@@ -114,12 +114,12 @@ macro_rules! single_value_type {
     ($trait_fn:ident, $visitor_fn:ident) => {
         fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
         where
-            V: Visitor<'de>
+            V: Visitor<'de>,
         {
             let v = parse_single_value(self.values)?;
             visitor.$visitor_fn(v)
         }
-    }
+    };
 }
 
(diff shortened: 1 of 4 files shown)


FSMaxB:
I really truly do not understand why the code coverage went down (or how to interpret the codecov output).

msrd0 (approved):
LGTM, @pksunkara @sezna what do you think?

pksunkara:
This is breaking and while I can approve it, I can not merge it without another maintainer reviewing it.

pksunkara (approved):

PR #430, opened 2020-05-13
Allow specifying the error type for NewMiddleware and related traits

msrd0:
I just tried to implement `NewMiddleware` for my own type and realised that I had an error that could occur but was not of type `std::io::Error`. Returning `io::Error::Other` along with a custom error message would work but I believe using the `?` shorthand is much easier.

This is why I introduced `NewMiddleware::Err` type to be specified manually. For most implementations, `std::convert::Infallible` is the appropriate error type (this is at least true for all types that use the derive macro), and those that wish to continue returning `io::Result` can simply specify `type Err = io::Error`. So this change does not hurt existing implementations but makes future ones much more convenient.

### DIFF
--- a/gotham/src/middleware/session/mod.rs
+++ b/gotham/src/middleware/session/mod.rs
@@ -1,6 +1,5 @@
 //! Defines a session middleware with a pluggable backend.
 
-use std::io;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 use std::panic::RefUnwindSafe;
@@ -16,6 +15,7 @@ use hyper::{Body, Response, StatusCode};
 use log::{error, trace, warn};
 use rand::RngCore;
 use serde::{Deserialize, Serialize};
+use thiserror::Error;
 
 use super::cookie::CookieParser;
 use super::{Middleware, NewMiddleware};
@@ -40,14 +40,17 @@ pub struct SessionIdentifier {
 }
 
 /// The kind of failure which occurred trying to perform a session operation.
-#[derive(Debug)]
+#[derive(Debug, Error)]
 pub enum SessionError {
     /// The backend failed, and the included message describes the problem.
+    #[error("The backend failed to return a session: {0}")]
     Backend(String),
     /// The session was unable to be deserialized.
+    #[error("The backend failed to deserialize the session")]
     Deserialize,
     /// Exhaustive match against this enum is unsupported.
     #[doc(hidden)]
+    #[error("__NonExhaustive")]
     __NonExhaustive,
 }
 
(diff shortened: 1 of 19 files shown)


msrd0:
Well, 0.5 can be a breaking change, so if you are happy with the API churn I don't see why this couldn't be merged. Unfortunately [associated type defaults](https://github.com/rust-lang/rust/issues/29661) aren't stable yet, otherwise we could make this change a non-breaking one.

gotham/src/middleware/session/mod.rs
```diff
@@ -40,14 +40,17 @@ pub struct SessionIdentifier {
...
 pub enum SessionError {
     /// The backend failed, and the included message describes the problem.
+    #[error("The backend failed to return a session: {0}")]
     Backend(String),
     /// The session was unable to be deserialized.
+    #[error("The backend failed to deserialize the session")]
     Deserialize,
     /// Exhaustive match against this enum is unsupported.
     #[doc(hidden)]
+    #[error("__NonExhaustive")]
```
pksunkara, line 53: You can simply specify `#[non_exhaustive]` on this enum if we bump MSRV to 1.40

pksunkara:
One thing I had recently tried to grapple with is that because of the associated types of `NewMiddleware`, we can't store the middlewares in an abstract form using trait objects. (which is why we needed the `borrow_bag` in the first place)

I understand that `NewMiddleware` returning another object as `Middleware` using `Instance` helps flexibility but I was wondering why they can't together be simply one trait without the `Instance` associated type.

To be honest, we don't need the `Err` associated type if we can specify an ECS type storage `gotham::Error` and use it for all middlewares like how [tide](https://docs.rs/tide/0.9.0/tide/struct.Error.html#method.new) does.

Basically, I am trying to make the usage middlewares DRY. Currently, it isn't.

msrd0:
Well, having an associated `Err` type is common practice in the std library, for example in the [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) or [`TryInto`](https://doc.rust-lang.org/std/convert/trait.TryInto.html) traits.

That said, if gotham doesn't really do anything with the error, especially not returning it to the user, I don't see a reason for not using `gotham::error::Error` (and possibly switch this from `failure`, which has been deprecated, to `anyhow`, which is the suggested successor).

However, since @nyarly prefers to break less code, I guess it is easier for now to be able to still use `std::io::Error`, and switch to a generic error type in the future?

pksunkara:
`this-error` is the suggested successor for libraries.

I don't understand why we don't want to do breaking changes. As long as they are documented, breaking stuff is always preferable because it will improve the library as a whole.

msrd0:
`anyhow` is the replacement of the generic error type while `thiserror` only provides the derive macro.

I though about this a bit further and I think that it actually makes more sense to allign `NewMiddleware` and `NewHandler` (which as of today returns `failure::Result`). Since the user does not get to handle the error, it makes no sense to use any concrete error type (including `std::io::Error`). This is why I created #438, which obsoletes this PR.

pksunkara:
Closing this PR in favor of #438

PR #437, opened 2020-05-23
Allow giving param names to globs thus allowing multiple

pksunkara:
@colinbankier @nyarly As you guys know, I had recently published [this blog post](https://pksunkara.com/posts/state-of-routing-in-rust/) comparing routing capabilities.

This was the only deficiency in gotham when compared to Rails router. So, I fixed it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,8 +82,3 @@ members = [
 
 
 ]
-
-[patch.crates-io]
-gotham = { path = "gotham" }
-gotham_derive = { path = "gotham_derive" }
-borrow-bag = { path = "misc/borrow_bag" }
--- a/gotham/Cargo.toml
+++ b/gotham/Cargo.toml
@@ -25,7 +25,7 @@ hyper = "0.13.1"
 serde = "1.0"
 serde_derive = "1.0"
 bincode = "1.0"
-mime = "0.3"
+mime = "0.3.15"
 # Using alpha version of mime_guess until mime crate stabilizes (releases 1.0).
 # see https://github.com/hyperium/mime/issues/52
 mime_guess = "2.0.1"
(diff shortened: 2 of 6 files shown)


gotham/Cargo.toml
```diff
@@ -25,7 +25,7 @@ hyper = "0.13.1"
 serde = "1.0"
 serde_derive = "1.0"
 bincode = "1.0"
-mime = "0.3"
+mime = "0.3.15"
```
pksunkara, line 28: I did this because we are using `Mime::essence_str()` which is not available in `<3.15`

Cargo.toml
```diff
@@ -82,8 +82,3 @@ members = [
 
 
 ]
-
-[patch.crates-io]
-gotham = { path = "gotham" }
-gotham_derive = { path = "gotham_derive" }
-borrow-bag = { path = "misc/borrow_bag" }
```
nyarly, line 89: Could you restore this, please? For development, it's certainly easier to use these together, rather than publishing individually.

nyarly:
Also, I'd quibble that the one feature eveyone overlooks in Rails is bi-directional routing. Although it's interface is overbroad, being able to use the `path_to` view helper _et al_ is extremely handy.

Cargo.toml
pksunkara: This is not needed anymore because Cargo supports both `path` and `version` in `dependencies` which I added in a previous PR. I missed one spot and I fixed it in this PR.

pksunkara:
> Also, I'd quibble that the one feature eveyone overlooks in Rails is bi-directional routing. Although it's interface is overbroad, being able to use the path_to view helper et al is extremely handy.

I agree but I don't understand why we are talking about this. Was there something related in this PR?

Cargo.toml
pksunkara: And since cargo doesn't support publishing all packages at once, you can use https://crates.io/crates/cargo-workspaces to do it which works with `path` and `version` in `dependencies` without this being needed.

pksunkara: @nyarly Can you please reply to this comment? I want to get this PR resolved.

nyarly:
> I agree but I don't understand why we are talking about this. Was there something related in this PR?

It's in reply to "This was the only deficiency in gotham when compared to Rails router." Not an issue with the change.

PR #438, opened 2020-05-28
Improve Error Handling

msrd0:
This Pull Request removes all usages of the now outdated `failure` crate and replaces them with `anyhow`, the suggested successor for a generic error type.

It also aligns the API of `NewMiddleware` with the API of `NewHandler`. Since the user does not get to handle the error types but all we do is generate an error response, we don't need a concrete error type like `std::io::Error` but can use a generic one. This simplifies the use of non-io errors since they can now use the shorthand `?` operator instead of going the considerably more elaborate `io::Error::Other` route.

This is a breaking API change but I expect the required changes to be minimal. Most users don't have many manual implementations of `NewHandler` or `NewMiddleware`, and for those implementations, all they have to do is change `io::Result` into `anyhow::Result` (which is re-exported by gotham).


msrd0:
@nyarly @colinbankier @pksunkara any updates on this?

msrd0:
**Update:** I have now also refactored `HandlerError` to have a `From<E>` implementation for all error types that implement `Into<anyhow::Error> + Display`. This allows to use all types that implement `std::error::Error` plus `anyhow::Error` itself as the error type for handlers, with `?` shorthand support being completed with #450. Also, for convenience, `Result<T, E>` and `TryFuture<T, E>` have a `map_err_with_status(status)` method that is basically equivalent to `map_err(|err| HandlerError::from(err).with_status(status))`.

CC @tanriol

sezna:
So there are two options to enable this -- we can either loosen or remove the code coverage requirement, or you can add more test coverage. Obviously the latter is preferred, but only if it is practical.

msrd0:
Afaik, code coverage is optional and merging is blocked because this pull request has no review (and also because there were merge conflicts which I just resolved). If you think some part of this PR is poorly tested, please let me know.

sezna:
I didn't realize the codecov check was non-blocking. I'll look at this.

sezna (dismissed):
LGTM!

msrd0:
I resolved another merge conflict, should be good for merge now @sezna

PR #446, opened 2020-06-23
Update comment in graceful shutdown example

zhming0:
This PR added some comments to the graceful shutdown example to avoid misunderstanding from user. 

Specifically, before the change, the example implies the server will get fully shut down before that `println`, but in reality, it may or may not be done yet.


msrd0:
I guess some clarification can't hurt, eventhough _gracefully_ implies to me that existing connections are not immediately closed but waited until they are done.

Can you fix the `rustfmt` issue?

sezna:
I'm going to close this -- feel free to reopen if you'd like to handle the fmt issues and discuss the wording.

Issue #434, opened 2020-05-19
Gotham's 405 response contradicts itself


msrd0:
I have several routes on the `/users` endpoint, including several `OPTIONS` handlers, each with a matcher looking for the `Access-Control-Request-Method` header, and each returning a 404 on route non match.

However, gotham decides to return a _405 Method Not Allowed_ response when all those matchers failed, saying that `OPTIONS` is not allowed, while listing `OPTIONS` in the `Allow` header:

![Screenshot_2020-05-19-14-12-15](https://user-images.githubusercontent.com/7853372/82325256-3851e480-99db-11ea-91fd-cd5c52c04969.png)

I don't know what is the appropriate return in this case, but I'd assume a _204 No Content_ with the `Allow` header set makes more sense.

Also, even more confusing, the `PATCH` method is included in the `Allow` header, although I do not have any handler registered to the router that would accept this method (and also don't have an option handler looking for `Access-Control-Request-Method: Patch`). So I don't think that `Allow: PATCH` should've been returned from gotham.

msrd0:
This should've been fixed by #447

PR #447, opened 2020-06-24
Improve status code and allow headers returned by RouteNonMatch based responses

msrd0:
This pull request improves the status code and `Allow`-headers return by gotham when no route was successful due to route matchers.

 - When two route matchers are combined with a logical AND and one of them returned 405, this error is preserved throughout this route. This change does not change those combined with a logical OR.
 - Routes combined with a logical OR now prefer 404 over 405 responses. This should fix the status-code part of #434.
 - The default `RouteNonMatch` includes no allowed methods, which prevents non-declared methods appearing as seen in #434. Therefore, the intersection logic was changed to preserve a method set if the other one is empty. I expect there to be no impact for well-written route matchers in user code. Overall, this change should make the behaviour clearer and more straight-forward, avoiding suprises when writing route matchers or even just using the ones provided with gotham.

### DIFF
--- a/gotham/src/router/non_match.rs
+++ b/gotham/src/router/non_match.rs
@@ -79,11 +79,11 @@ impl RouteNonMatch {
                 StatusCode::METHOD_NOT_ALLOWED
             }
             // For 404, prefer routes that indicated *some* kind of match.
-            (StatusCode::NOT_FOUND, rhs) => rhs,
-            (lhs, StatusCode::NOT_FOUND) => lhs,
+            (StatusCode::NOT_FOUND, rhs) if rhs.is_client_error() => rhs,
+            (lhs, StatusCode::NOT_FOUND) if lhs.is_client_error() => lhs,
             // For 406, allow "harder" errors to overrule.
-            (StatusCode::NOT_ACCEPTABLE, rhs) => rhs,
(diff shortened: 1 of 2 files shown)


gotham/src/router/non_match.rs
```diff
@@ -86,7 +102,23 @@ impl RouteNonMatch {
     /// which wrap other `RouteMatcher` instances. See the `Node::select_route` implementation (in
     /// `gotham::router::tree`) for an example.
     pub fn union(self, other: RouteNonMatch) -> RouteNonMatch {
-        let status = higher_precedence_status(self.status, other.status);
+        let status = match (self.status, other.status) {
```
sezna, line 105: I guess this can't just be an `impl PartialOrd` because `StatusCode` originates from Hyper? If we ever want to add any more gotham specific functionality, it could be worth a newtype, at which point this would be simplified.

msrd0: The point of this PR is to use different compare logic for union vs intersection, so `impl PartialOrd` wouldn't do it either way.

sezna: Oh, my eyes glazed over the fact that they are different. LGTM!

PR #449, opened 2020-06-29
WIP: "closure implements `Fn`, so references to captured variables ca…

alsuren:
…n't escape the closure"

### DIFF
--- a/gotham/src/router/builder/single.rs
+++ b/gotham/src/router/builder/single.rs
@@ -1,10 +1,10 @@
-use hyper::Body;
+use hyper::{Body, Response};
 
 use std::panic::RefUnwindSafe;
 
 use crate::extractor::{PathExtractor, QueryStringExtractor};
 use crate::handler::assets::{DirHandler, FileHandler, FileOptions, FilePathExtractor};
-use crate::handler::{Handler, HandlerResult, NewHandler};
+use crate::handler::{Handler, HandlerError, HandlerResult, NewHandler};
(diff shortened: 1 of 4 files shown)


gotham/src/router/builder/single.rs
```diff
@@ -157,6 +177,29 @@ pub trait DefineSingleRoute {
...
+                    let result = fut.await;
+                    match result {
+                        Ok(response) => Ok((state, response)),
+                        Err(err) => Err((state, err)),
+                    }
+                }
+                .boxed()
+            })
+        })
+    }
```
alsuren, line 202:
```
error: lifetime may not live long enough
   --> gotham/src/router/builder/single.rs:191:17
    |
190 |               Ok(move |state: State| {
    |                  -------------------
    |                  |                 |
    |                  |                 return type of closure is std::pin::Pin<std::boxed::Box<(dyn core::future::future::Future<Output = std::result::Result<(state::State, http::response::Response<hyper::body::body::Body>), (state::State, handler::error::HandlerError)>> + std::marker::Send + '2)>>
    |                  lifetime `'1` represents this closure's body
191 | /                 async {
192 | |                     let fut = handler.call(&state);
193 | |                     let result = fut.await;
194 | |                     match result {
...   |
198 | |                 }
199 | |                 .boxed()
    | |________________________^ returning this value requires that `'1` must outlive `'2`
    |
    = note: closure implements `Fn`, so references to captured variables can't escape the closure
```

PR #450, opened 2020-06-29
Async borrowing (continuation of #449)

tanriol:
It seems to work as expected.

Todo:
- [x] Try to avoid mapping every error with `into_handler_error` before `?`

Feedback needed:
- [x] Could this allow for generic return types like `R: IntoResponse`? Yes.
- [x] Could this allow for handlers receiving either `&State` or `&mut State`? Not really.
- [x] Could this allow for returning the reponse without Ok-wrapping in the infallible case? Not really.

And "do we want that" for every question above :-)

### DIFF
--- a/gotham/src/router/builder/single.rs
+++ b/gotham/src/router/builder/single.rs
@@ -23,16 +23,16 @@ trait FnHelper<'a> {
         + Send
         + Sync
         + 'a;
-    fn call(self, arg: &'a State) -> Self::Fut;
+    fn call(self, arg: &'a mut State) -> Self::Fut;
 }
 
 impl<'a, D, F> FnHelper<'a> for F
 where
-    F: FnOnce(&'a State) -> D,
+    F: FnOnce(&'a mut State) -> D,
     D: std::future::Future<Output = Result<Response<Body>, HandlerError>> + RefUnwindSafe + Copy + Send + Sync + 'a,
 {
     type Fut = D;
-    fn call(self, state: &'a State) -> D {
+    fn call(self, state: &'a mut State) -> D {
         self(state)
     }
 }
@@ -187,9 +187,9 @@ pub trait DefineSingleRoute {
         // Fut: Future<Output = HandlerResult> + Send + 'static
     {
         self.to_new_handler(move || {
-            Ok(move |state: State| {
+            Ok(move |mut state: State| {
                 async {
-                    let fut = handler.call(&state);
+                    let fut = handler.call(&mut state);
                     let result = fut.await;
                     match result {
                         Ok(response) => Ok((state, response)),
--- a/gotham/src/router/builder/single.rs
+++ b/gotham/src/router/builder/single.rs
@@ -1,10 +1,13 @@
 use hyper::Body;
 
 use std::panic::RefUnwindSafe;
+use std::pin::Pin;
 
 use crate::extractor::{PathExtractor, QueryStringExtractor};
 use crate::handler::assets::{DirHandler, FileHandler, FileOptions, FilePathExtractor};
-use crate::handler::{Handler, HandlerError, HandlerResult, IntoResponse, NewHandler};
+use crate::handler::{
+    Handler, HandlerError, HandlerFuture, HandlerResult, IntoResponse, NewHandler,
+};
 use crate::pipeline::chain::PipelineHandleChain;
 use crate::router::builder::{
     ExtendRouteMatcher, ReplacePathExtractor, ReplaceQueryStringExtractor, SingleRouteBuilder,
@@ -15,6 +18,10 @@ use crate::router::route::{Delegation, Extractors, RouteImpl};
 use crate::state::State;
 use core::future::Future;
 use futures::FutureExt;
+
+pub trait HandlerMarker {
+    fn call_and_wrap(self, state: State) -> Pin<Box<HandlerFuture>>;
+}
 
 pub trait AsyncHandlerFn<'a> {
     type Res: IntoResponse + 'static;
@@ -34,6 +41,28 @@ where
         self(state)
     }
 }
+
+impl<F, R> HandlerMarker for F
+where
+    R: IntoResponse + 'static,
+    for<'a> F: AsyncHandlerFn<'a, Res = R> + Send + 'static,
+{
+    fn call_and_wrap(self, mut state: State) -> Pin<Box<HandlerFuture>> {
+        async move {
+            let fut = self.call(&mut state);
+            let result = fut.await;
+            match result {
+                Ok(data) => {
+                    let response = data.into_response(&state);
+                    Ok((state, response))
+                }
+                Err(err) => Err((state, err)),
+            }
+        }
+        .boxed()
+    }
+}
+
 /// Describes the API for defining a single route, after determining which request paths will be
 /// dispatched here. The API here uses chained function calls to build and add the route into the
 /// `RouterBuilder` which created it.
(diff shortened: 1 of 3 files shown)


gotham/src/router/builder/single.rs
```diff
@@ -16,6 +16,24 @@ use crate::state::State;
...
+    type Res: IntoResponse + 'static;
+    type Fut: std::future::Future<Output = Result<Self::Res, HandlerError>> + Send + 'a;
+    fn call(self, arg: &'a mut State) -> Self::Fut;
+}
+
+impl<'a, D: 'a, R, F> AsyncHandlerFn<'a> for F
+where
+    F: FnOnce(&'a mut State) -> D,
+    R: IntoResponse + 'static,
+    D: std::future::Future<Output = Result<R, HandlerError>> + Send + 'a,
```
technic, line 29: The `D: 'a` requirement is duplicated in the line 25

tanriol: Removed it from line 25.

gotham/src/lib.rs
```diff
@@ -3,8 +3,8 @@
 //! You can find out more about Gotham, including where to get help, at <https://gotham.rs>.
 //!
 //! We look forward to welcoming you into the Gotham community!
-#![doc(html_root_url = "https://docs.rs/gotham/0.5.0-rc.1")] // Update when changed in Cargo.toml
-#![warn(missing_docs, deprecated)]
+#![doc(html_root_url = "https://docs.rs/gotham/0.5.0-rc.1")]
+// Update when changed in Cargo.toml
```
alsuren, line 7: This was just me trying to remove the noise from my `cargo watch` output. Probably revert this and fix it in a separate PR.

gotham/src/router/builder/single.rs
```diff
@@ -16,6 +16,24 @@ use crate::state::State;
...
+    type Res: IntoResponse + 'static;
+    type Fut: std::future::Future<Output = Result<Self::Res, HandlerError>> + Send + 'a;
+    fn call(self, arg: &'a mut State) -> Self::Fut;
+}
+
+impl<'a, D, R, F> AsyncHandlerFn<'a> for F
+where
+    F: FnOnce(&'a mut State) -> D,
+    R: IntoResponse + 'static,
+    D: std::future::Future<Output = Result<R, HandlerError>> + Send + 'a,
```
alsuren, line 29: this was only called D because the original gist was trying to return `impl Debug + 'a`. Probably want some better names for this.

gotham/src/service/trap.rs
```diff
@@ -72,8 +72,8 @@ fn finalize_error_response(
         // API doesn't guarantee its presence (even though it always is).
         let err_description = err
             .source()
-            .map(Error::description)
-            .unwrap_or_else(|| err.description());
+            .map(Error::to_string)
+            .unwrap_or_else(|| err.to_string());
```
alsuren, line 76: Let's revert this too, and put it in its own PR.

gotham/src/router/builder/single.rs
```diff
@@ -157,6 +175,31 @@ pub trait DefineSingleRoute {
...
+                    let result = fut.await;
+                    match result {
+                        Ok(data) => {
+                            let response = data.into_response(&state);
+                            Ok((state, response))
+                        }
+                        Err(err) => Err((state, err)),
+                    }
+                }
+                .boxed()
```
alsuren, line 199: is it possible to move this up into the AsyncHandlerFn as a `call_wrapped(State) -> Box<dyn Future<Output = Result<(State, _), (State, _)>>`? At the moment, there are a lot of complex moving parts that are spread all over the file.

gotham/src/lib.rs
tanriol: Should I revert these or remove these changes when rebasing the patchset? I'm going to rebase my part anyway, but not sure whether I should touch your commit too or leave it as-is.

gotham/src/router/builder/single.rs
tanriol: It may be possible to move this into a free function or another trait, but it explicitly cannot be moved into `AsyncHandlerFn` trait itself because it uses HRTB on that.

gotham/src/lib.rs
alsuren: I'm happy for you to keep it as a separate commit or fixup my commit if you want. Up to you.

tanriol:
I've tried making this work for an async function of `&State`. The problem is that for the future to be `Send` we'd need `&State` to be `Send`, which, according to the standard rules for shared references, requires `State` to be `Sync`.

tanriol:
Looks like this would not work without `Result` in the infallible case because of ambiguous error type. The problem here is that there's the following impl
```rust
impl<T, E> IntoResponse for Result<T, E> where
    T: IntoResponse,
    E: IntoResponse, 
{ ... }
```
which means that returning `impl IntoResponse` and passing the errors with `?` will result in Rust being unable to infer the `E` type from the `ErrorTypeN: Into<E>` and `E: IntoResponse` bounds.

technic:
As far as I understand rust uses `From` trait for converting between types when `?` operator is used (or `Into`). 
So if the want to go via intermediate type which implements `IntoResponse` there is no way for rust to deduce this type automatically.

tanriol:
Pushed an updated version. I haven't looked into `IntoHandlerError` because @msrd0 said that may go in #438.

The main question, IMHO, is whether we want to keep both `to_async` and `to_async_borrowed` or just one of them?

tanriol:
@technic 
> As far as I understand rust uses `From` trait for converting between types when `?` operator is used (or `Into`).
> So if the want to go via intermediate type which implements `IntoResponse` there is no way for rust to deduce this type automatically.

Not exactly. Rust can handle inference like "there's an Ok being returned, so this is a Result, and the only impl for a Result is an impl for `Result<_, HandlerError>`, so the error type is known to be a HandlerError". If the impl above were an `impl<T: IntoResponse> IntoResponse for Result<T, HandlerError>`, it would compile successfully.

msrd0:
What is the status of this pull request? #438 has finally been merged, so this PR should no longer be blocked on it.

tanriol:
Will rebase in a couple days (maybe faster if the rebase is easy). IIRC, there's the problem that two different variants allow either borrowing (and passing an async function) in async or passing a closure (but no borrowing), not both.

alsuren:
I will admit that I've stepped away from both of these projects, but I occasionally get notifications from both projects, so I'm leaving this here in case it provides inspiration for anyone who comes along.

The people over at the goose project seem to have made some headway with closures. Last I saw, they ask the user to wrap the closure in an Arc and Box::pin() the resulting future themselves (the Arc is goose-specific, because of how the containing struct derives Clone. Gotham could probably use Box there too. See https://github.com/tag1consulting/goose/pull/120/files#diff-1cd6d8660ef657deb6ee2a73b4d592d9R1922-R1926 https://github.com/tag1consulting/goose/pull/120/files#diff-8484c6c853111ca2278cc1dc616df376R189-R195 and https://github.com/tag1consulting/goose/pull/126/files) but there might have been some ergonomics improvements since then that I've not been keeping on top of (I'm not sure whether this approach went anywhere, for example: https://github.com/tag1consulting/goose/pull/94#issuecomment-665741431 . I suspect not). 

I suspect that supporting closures is more interesting for goose than it is for gotham. If the ergonomics of some theoretical `.to_borrowing_closure()` are a bit less nice than the owning .to() and `.to_async()` variants then maybe that's fine. It might be an argument for keeping the old variants around though, until rustc is capable of letting you express the appropriate types and lifetimes for your function arguments directly.

sezna (approved):

msrd0 (approved):

technic:
Great job! thanks!

Issue #467, opened 2020-09-04
Pass State into the Session Backend


FungusHumungus:
I am working on a Redis backend for storing the session. 

The way to do this, I imagine, is to create a separate Middleware to manage a connection pool to Redis. This Middleware would store a connection in the State that the Session backend can use. I don't see any other way that the Redis connections could be handled, correct me if I'm wrong.

In order for the backend to use the connection it needs to have access to the State. Currently, this isn't passed in. It would be great if it could be!

I have knocked up a quick proof of concept here: https://github.com/FungusHumungus/gothamboogy/blob/master/src/redis.rs, and have a tentative pull request available - if you think this is the correct way to solve this issue. I'll submit it in a minute for your perusal.

msrd0:
I just took a look at your proof of concept and I don't think I fully understand why you want your redis connection in the `State`. You don't seem to store a connection pool of any kind in either the middleware or the session backend, but create a new connection for every request. Even worse, you return a 500 if redis is not available before the handler even requests a session (which might be fine for your use case if every handler is guaranteed to require a session).

If I understand your description correctly, what you want is a connection pool for your session backend. Is there any reason why it is not possible for you to just store that in the backend but have to do so in a middleware? I'm not against passing the state to the session backend, just trying to understand why it's necessary.

FungusHumungus:
So I just haven't gone to the trouble of setting up a pool in the POC, it's just easier to create a new connection every time - but ultimately that should be handled by a proper pool. 

I agree, it's not ideal that to have a Redis backed session store, you have to then also create the Redis middleware, and horrible things would happen at runtime if you forgot to add the Redis middleware. But since (in my case anyway) the handlers will likely also want a Redis connection, it makes sense to use the same connection for the session and the handlers.

The alternative, as you suggested to create the pool and store it in the backend would certainly be possible. I suspect though, to get a connection from the pool would probably involve a future - as it does in [darkredis](https://docs.rs/darkredis/0.7.0/darkredis/struct.ConnectionPool.html#method.get). In which case the backend methods (`persist_session` and `drop_session`) would presumably also need to return a future.

I would be happy to have a go at getting that to work if you think it would be a better solution, or if you can think of any other better ideas?

msrd0:
Ok I see so you're trying to have one redis connection pool that can be shared with the session backend and the handlers. In that case it probably makes sense to use a middleware, eventhough you'll need to ensure that the redis middleware will always run before the session middleware.

[...]

PR #468, opened 2020-09-04
Send the state to the Session backend.

StephenWakely:
This is for #467.

It sends the State to the `persist_session`, `read_session` and `drop_session` methods of the Backend. This will be useful for additional backends, such as a Redis backend, that will need to access a connection that will be stored in the State.


msrd0:
Thanks for the PR!

Can you please fix the issues reported by [travis](https://travis-ci.com/github/gotham-rs/gotham/builds/182785404) - that is, the test code needs to be updated and after that you'll need to run `rustfmt`.

FungusHumungus:
I've now modified it so the Backend functions returned futures. This should allow for more flexibility when defining different backends.

gotham/src/middleware/session/backend/mod.rs
```diff
@@ -27,17 +31,26 @@ pub trait Backend: Send {
     /// Persists a session, either creating a new session or updating an existing session.
     fn persist_session(
         &self,
+        state: &State,
         identifier: SessionIdentifier,
         content: &[u8],
-    ) -> Result<(), SessionError>;
+    ) -> Pin<Box<SetSessionFuture>>;
```
msrd0, line 37: Sorry for taking so long to review this - I believe this type signature is incorrect? Without any lifetime information, the `GetSessionFuture` type needs to live for `'static`, which it will not do if it accessed the state in an async context. I think this needs to be changed to something like

```rust
persist_session<'fut, 'a: 'fut, 'b: 'fut, 'c: 'fut>(
    &'a self,
    state: &'b State,
    identifier: SessionIdentifier,
    content: &'c [u8]
) -> Pin<Box<GetSessionFuture + 'fut>>
```

Alternatively, it might make sense to let `#[async_trait]` deal with the lifetimes.

gotham/src/middleware/session/backend/mod.rs
```diff
@@ -27,17 +31,26 @@ pub trait Backend: Send {
...
     ///
     /// The returned future will resolve to an `Option<Vec<u8>>` on success, where a value of
     /// `None` indicates that the session is not available for use and a new session should be
     /// established.
-    fn read_session(&self, identifier: SessionIdentifier) -> Pin<Box<SessionFuture>>;
+    fn read_session(
+        &self,
+        state: &State,
+        identifier: SessionIdentifier,
+    ) -> Pin<Box<GetSessionFuture>>;
```
msrd0, line 48: Same as above

gotham/src/middleware/session/backend/mod.rs
```diff
@@ -27,17 +31,26 @@ pub trait Backend: Send {
...
 
     /// Drops a session from the underlying storage.
-    fn drop_session(&self, identifier: SessionIdentifier) -> Result<(), SessionError>;
+    fn drop_session(
+        &self,
+        state: &State,
+        identifier: SessionIdentifier,
+    ) -> Pin<Box<SetSessionFuture>>;
```
msrd0, line 55: Same as above

msrd0 (changes requested):

gotham/src/middleware/session/backend/mod.rs
FungusHumungus: No worries, thanks for reviewing. I'll try to get to these shortly.

The last time I used `#[async_trait]` it resulting in the rust compiler not being able to produce any decent error messages for any errors in the code that implemented those traits. We ended up changing back to returning Futures instead. Do you know if this problem has been resolved? I would be hesitant to use it unless it has..

msrd0: The only time I used `#[async_trait]` was to run `cargo expand` on it to see how it worked because I didn't want my proc macro to invoke another proc macro. So I don't know if this has been changed or has ever been a problem. However, I feel like the code would probably be more readable if we don't introduce this many lifetimes, but if you run into any incorrect/unhelpful error messages, just write the code the explicit way. I'm happy with both solutions.

msrd0:
Do you think you can get this PR ready? I'd like to include it in the next release. Let me know if you need any help.

StephenWakely:
Hi, really sorry, I haven't had the chance to get to this. Life has taken over. I don't think I will get the chance to look at it any time soon either. I'm happy for someone else to take this on if they are willing, or we can just leave it and I will get to at it some point, whenever that may be!

msrd0 (approved):
No problem. I'm going to merge this as-is as it should work for most use cases, e.g. pulling a database connection out of the state and then enter an async block in which that connection is being used, and we can always improve it in the future.

PR #475, opened 2020-09-11
Allow downcasting of the HandlerError's cause

### LINKED ISSUE #135: Downcasting of `HandlerError` cause.
It would be great if the `HandlerError` had an interface similar to that of `failure::Error`, specifically the ability to wrap the underlying `Error` such that it can be downcast: https://boats.gitlab.io/failure/error.html#downcasting

This would be especially useful in cases where, in our gotham handlers & middleware, we simply map various error types into `HandlerError` via `into_handler_error()`. This would then allow for the creation and use of a middleware dedicated to error handling **AS WELL AS** the ability to then downcast that error and determine if it can be directly serialized into JSON (or some other content-type) for display to the user.

E.G., one of our handlers needs to return a validation error to the user. A specific error type would be used for this which can be directly serialized to JSON and then sent to a user. We would want to simply propagate that error up through the middleware pipeline & then have one uniform location for performing serialization of errors flowing up the chain, and also recognizing if it is an internal error which should be represented as an opaque 500 to the user &c.

Definitely interested in some discussion on this.

msrd0:
Fixes #135 by exposing `anyhow`'s downcasting API.

### DIFF
--- a/gotham/Cargo.toml
+++ b/gotham/Cargo.toml
@@ -52,6 +52,7 @@ tokio-rustls = { version = "0.14.0", optional = true }
 
 [dev-dependencies]
 gotham_derive = { path = "../gotham_derive" }
+thiserror = "1.0"
 
 [badges]
 travis-ci = { repository = "gotham-rs/gotham", branch = "master" }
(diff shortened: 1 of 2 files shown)


nyarly:
LGTM, once the rustfmt is fixed.

msrd0:
fixed

nyarly (approved):

PR #478, opened 2020-10-03
Update Tera example, macro compile_templates! is deprecated

### LINKED ISSUE #464: Update the tera example
The tera example uses the `compile_templates!` macro that has been removed in version 1.0 ([Changelog](https://github.com/Keats/tera/blob/master/CHANGELOG.md#100-2019-12-07)). It needs to be rewritten to compile with newer versions of tera.

Related: #460

edjroz:
Update Tera example to comply with Tera 1.5

Close #464

### DIFF
--- a/examples/templating/tera/src/main.rs
+++ b/examples/templating/tera/src/main.rs
@@ -1,19 +1,22 @@
 //! An example usage of Tera template engine working with Gotham.
 #[macro_use]
 extern crate lazy_static;
-#[macro_use]
-extern crate tera;
 
 use gotham::state::State;
 use tera::{Context, Tera};
 
-// Assuming the Rust file is at the same level as the templates folder
-// we can get a Tera instance that way:
 lazy_static! {
-    pub static ref TERA: Tera =
-        compile_templates!(concat!(env!("CARGO_MANIFEST_DIR"), "/templates/**/*"));
+    pub static ref TERA: Tera = {
+        let tera = match Tera::new(concat!(env!("CARGO_MANIFEST_DIR"), "/TEMPLATES/**/*")) {
+            Ok(t) => t,
+            Err(e) => {
+                println!("Parsing error(s): {}", e);
+                ::std::process::exit(1);
+            }
+        };
+        tera
+    };
 }
-
 /// Create a `Handler` which calls the Tera static reference, renders
 /// a template with a given Context, and returns the result as a String
 /// to be used as Response Body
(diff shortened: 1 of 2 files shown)


examples/templating/tera/src/main.rs
```diff
@@ -1,19 +1,22 @@
...
-    pub static ref TERA: Tera =
-        compile_templates!(concat!(env!("CARGO_MANIFEST_DIR"), "/templates/**/*"));
+    pub static ref TERA: Tera = {
+        let tera = match Tera::new(concat!(env!("CARGO_MANIFEST_DIR"), "/TEMPLATES/**/*")) {
+            Ok(t) => t,
+            Err(e) => {
+                println!("Parsing error(s): {}", e);
+                ::std::process::exit(1);
+            }
+        };
```
msrd0, line 16: Can't we just use `.expect` instead of the `match` block here?

msrd0 (changes requested):
Hi, thanks for the PR! The travis test failed, can you fix that please?

examples/templating/tera/src/main.rs
edjroz: We could but Ideally wouldn't we want to keep in on par with Tera's documentation?

msrd0: I think improving readability is more important than using code from Tera's documentation

examples/templating/tera/src/main.rs
```diff
@@ -1,19 +1,16 @@
...
 
 use gotham::state::State;
 use tera::{Context, Tera};
 
-// Assuming the Rust file is at the same level as the templates folder
-// we can get a Tera instance that way:
 lazy_static! {
-    pub static ref TERA: Tera =
-        compile_templates!(concat!(env!("CARGO_MANIFEST_DIR"), "/templates/**/*"));
+    pub static ref TERA: Tera = {
```
msrd0, line 9: I believe this block can now be made into a single expression

edjroz:
> Hi, thanks for the PR! The travis test failed, can you fix that please?

are you able to replicate this error locally ? I'm running on Darwin and can't seem to replicate it by running `cargo test -j2 --all`

msrd0:
Yes, I can replicate this issue on ArchLinux. I noticed you changed the template folder name, maybe you should double-check if that was actually intended?

msrd0 (approved):

PR #488, opened 2020-10-23
WIP: remove StateData trait

### LINKED ISSUE #487: RFC: remove StateData trait
Currently to be able to store something in `State`, the data needs to implement the `StateData` trait. This `StateData` trait is really just an alias for `Any + Send`. It would be really nice if implementing this trait would not be needed. Here are some reasons why:

1. **Not technically needed**. This abstraction is not actually needed. As you can see in [this branch](https://github.com/lucacasonato/gotham/tree/remove-statedata), the `StateData` can be removed and replaced with direct `Any + Send` without issues. There is thus no technical reason for the existence of this trait.
2. **Unnecessary boilerplate**. To be able to store a struct in `gotham::state::State`, you need to either add the `StateData` proc-macro to the declaration, or manually implement `gotham::state::StateData`. This is an extra line of code that is not needed.
3. **Forces unnecessary wrapper structs**. If you want to store a structure from a crate outside of yours, you need to wrap it with of a wrapper element just so you can implement `StateData`. This adds unnecessary bloat.

I propose that the `StateData` trait and `StateData` proc-macro are removed. All methods on `State` and `FromState` that currently take a `StateData` would take any `Any + Send` instead.

This would also be one less thing that users need to learn about when starting out with Gotham.

lucacasonato:
Closes #487.

This PR is still WIP, because documentation has not been updated yet. If this RFC is accepted, I will update the related docs.


examples/handlers/multipart/src/main.rs
```diff
@@ -15,7 +15,7 @@ use std::pin::Pin;
 /// Extracts the elements of the POST request and responds with the form keys and values
 fn form_handler(mut state: State) -> Pin<Box<HandlerFuture>> {
     const BOUNDARY: &str = "boundary=";
-    let header_map = HeaderMap::take_from(&mut state);
+    let header_map = HeaderMap::<HeaderValue>::take_from(&mut state);
```
msrd0, line 18: According to hyper's docs, the type param for `HeaderMap` has `HeaderValue` as its default so this explicit form shouldn't be necessary

lucacasonato: Yeah I thought so too. But for some reason rust can not infer the type:

```
error[E0282]: type annotations needed for `gotham::hyper::HeaderMap<T>`
  --> examples/handlers/multipart/src/main.rs:20:10
   |
18 |     let header_map = HeaderMap::take_from(&mut state);
   |         ---------- consider giving `header_map` the explicit type `gotham::hyper::HeaderMap<T>`, where the type parameter `T` is specified
19 |     let boundary = header_map
20 |         .get(CONTENT_TYPE)
   |          ^^^ cannot infer type for type parameter `T`
   |
   = note: type must be known at this point

error: aborting due to previous error; 1 warning emitted

For more information about this error, try `rustc --explain E0282`.
error: could not compile `gotham_examples_handlers_multipart`.
```

msrd0: Oh I see, `let header_map: HeaderMap = HeaderMap::take_from(&mut state)` does work, so the problem is that `Any + Send` is also implemented for other types `T` whereas `StateData` was only implemented for the default `T`.

I dislike that you now need to write more code, but I also don't really know how to fix that. Any ideas?

lucacasonato: Ah right... so the same thing would have happened previously had you implemented `StateData` for `HeaderMap<T>` with `T` != `HeaderValue` (which you couldn't do previously from outside the `gotham` crate). So maybe this is actually a feature, not a bug. Being constrained to only the default `HeaderMap` seems arbitrary.

Also I think most users (including beginners) will know how to solve this error already (from cases like `.iter().map(...).collect()`).

msrd0: It's not arbitrary as that type is put into the state by gotham, so you wouldn't use any other `T`. Solving this error is easy but introduces unnecessary boilerplate, and also breaks existing code, and reading http headers is not that uncommon I'd assume.

msrd0: I just noticed that `let header_map: HeaderMap = state.take()` does also work and is shorter than what we currently have, so if we consistently advertised that API I might be ok with that.

lucacasonato: Good one. I have updated the examples to use `state.take()` / `state.borrow()`.

PR #496, opened 2020-11-03
Add a PathTemplate derive

msrd0:
This PR aims at replacing the current path extractor logic with a new, safer `PathTemplate` derive. This allows us to ensure that the template string and the path extracting struct match each other, so no runtime errors can occur.

## TODO

- [x] Add a `PathTemplate` trait that returns the template string
- [ ] Verify the template string against the input struct fields
- [ ] Create a struct instance from the request URI and put it into the state
- [ ] Update documentation and examples

## Open Questions

- [ ] Do we want to continue using serde or do we want our own logic?

### DIFF
--- a/gotham/src/router/builder/draw.rs
+++ b/gotham/src/router/builder/draw.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::marker::PhantomData;
 use std::panic::RefUnwindSafe;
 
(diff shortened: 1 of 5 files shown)


msrd0:
@nyarly Do you think we should keep using `serde::Deserialize` for path extractors or replace that entirely with our own derive logic?

tanriol:
Do I understand correctly that the TODO point about adding the structure to State effectively deprecates `with_path_extractor` and the `PathExtractor` trait?

msrd0:
I guess deprecating `PathExtractor` was my plan, but this is still open for discussion.

msrd0:
Alternatively, we could just verify that the type fields matches the url and have `.get_path()` just be a shortcut for `.get().with_path_extractor()`. That'd require us to continue using serde.

nyarly:
@msrd0 Did you want to start a Github discussion about this?

msrd0:
@nyarly I've never used github discussions before. I think the problem was that while I would prefer a more statically typed version than the current string-template path with a (hopefully) matching deserializable struct, I didn't really have a good idea of the alternative I tried to implement in this PR (that's why I closed it previously). Feel free to start a github discussion for this.

PR #503, opened 2020-11-29
Implement Into<Response> for the TestResponse

FSMaxB:
This makes working with the `TestResponse` much easier to work with if direct access to the underlying `Response` is required.


msrd0:
Thanks for the PR! Can you update the documentation for `TestResponse`? It currently suggests that only deref-ing is possible.

FSMaxB:
> Thanks for the PR! Can you update the documentation for TestResponse? It currently suggests that only deref-ing is possible.

Done

msrd0 (approved):

PR #511, opened 2020-12-24
Update hyper to 0.14 and tokio/bytes to 1.0

msrd0:
This PR updates hyper to 0.14 and tokio and bytes to 1.0.

~The websocket example has been disabled as tokio-tungstenite has not yet been updated (PR: https://github.com/snapview/tokio-tungstenite/pull/142).~ The websocket example has also been updated. Unfortunately, hyper no longer accepts upgrades on the `Body` type but requires `OnUpgrade` from the http requests extensions, so I put that type into the state whenever it appears in the request.

Note: hyper split up their code behind several feature flags. As of right now, gotham basically needs all features, since we have built-in test functionality that also depends on the client side of hyper. We might want to also expose feature flags that only depend on the hyper client code if we enabled our test code.


nyarly (approved):
Yay websockets

krallin:
Would it make sense to just expose the Extensions on the state directly, given that this is what Gotham does for the other parts that come from Hyper? 

For context, the reason I'm asking is because I'd like to send a PR to expose those, but if I'm exposing extensions, it's a little awkward if this one isn't there. Given this version hasn't been released yet, maybe there's time to reconsider this decision? :)

msrd0:
We could just put the extensions into the state directly. Since there seems to be no way of getting an iterator over all extensions, this might be the only way for now, especially for non-standard extensions that gotham can't know about. However, it feels a bit weird since both `State` and `Extensions` are basically wrappers over `HashMap<TypeId, Box<dyn Any>>`. 

The only alternative I can think of is providing a method that allows the user to process extensions and put them into the state - which I'm not a fan of since it'd involve a second type of middleware that'd probably just make stuff more complicated.

PR #518, opened 2021-01-08
Remove mio dependency

krallin:
It does not appear to be used at all, let's remove it? :)


msrd0:
This is a duplicate of #490

krallin:
Whoops, my bad, I rebased on the latest release and didn’t notice this has
been done already. Thanks!

Issue #183, opened 2018-03-17
Composible Service


nyarly:
In #182 I was saying

> In a perfect world, Gotham would provide convenience start and start_tls functions, as well as a way to get a tokio_service::Service so that it could be passed to other tokio consumers (especially hyper)

Really, the core to this idea is allowing a Gotham structure to be used in a composed HTTP service provider. #182 is one use case (in effect decorating the HTTP service in the TLS protocol), but another conceivable use case would be a Rust analogue of https://golang.org/pkg/net/http/pprof/ - which would involve muxing a Gotham service with another hyper client library.

While I appreciate that GothamService as it stands is probably too large an interface to export, perhaps there could be a `tokio_service::Service` built from or wrapping a `GothamService`?

The salt to take this with is that while I'm pretty sure that would solve my immediate use case ("use with `tokio_proto` and `native_tls`") I'm not completely clear about the "mux" case.

smangelsdorf:
This ties in with #95, in that both issues could be resolved with the same solution. I came up with a [rough idea](https://github.com/gotham-rs/gotham/issues/95#issuecomment-361544427) on how to make further progress there, but it will need a little more thought/experimentation before it suits this ticket as well.

@nyarly Does that get most of the way to what you're trying to achieve? Do you want to work on this?

The biggest gap that comes to mind is the inability to identify which `GothamListener` (to use the naming from my comment) created a connection, for example when http+https apps want to know whether a connection was secure or not.

nyarly:
I think your comments there are in the same vein, for sure. Lemme take a swing at it, sure.

cite-reader:
I took a look at this recently so I could plug a Gotham application into a Unix socket. The resulting branch is [here](https://github.com/cite-reader/gotham/tree/expose-service).

The tricky part is that the Service trait is only implemented for `ConnectedGothamService`, but the key feature of that struct is the `std::net::SocketAddr`. Transports that aren't IP 4 or 6 don't have that. My solution was to pull that into `GothamService` as an optional field, but I think some additional discussion is warranted before I go for a PR.

msrd0:
This has been mostly implemented in #526, but the client address is still required. @cite-reader I feel like unix sockets could easily be implemented by faking an address, e.g. always setting it to `::1` or something similar. Would that work for you?

cite-reader:
[...]

PR #526, opened 2021-02-24
state: move instantiation into State::from_request

krallin:
This refactors the service implementation a little bit to instantiate
the State via a dedicated method. The underlying goal is to allow users
to use Gotham without having to use its handler.

The motivation for this is to let users inject more things into their 
state and call their handler directly instead of having to go through a
Hyper service. In our case, we use this for:

- Integrating in a Hyper service for a sub-route.
- Injecting data about TLS client certificates that were used by the client

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,5 +81,7 @@ members = [
     # finalizer
     "examples/finalizers/",
 
+    # custom_service
+    "examples/custom_service/",
 
 ]
--- a/examples/custom_service/Cargo.toml
+++ b/examples/custom_service/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "gotham_examples_custom_service"
+description = "Using Gotham in a pre-existing Hyper service"
+version = "0.0.0"
+publish = false
+edition = "2018"
+
+[dependencies]
+anyhow = "1.0"
+futures = "0.3"
+gotham = { path = "../../gotham" }
+http = "0.2"
+hyper = { version = "0.14", features = ["full"] }
+tokio = { version = "1.0", features = ["full"] }
--- a/examples/custom_service/README.md
+++ b/examples/custom_service/README.md
@@ -1,0 +1,19 @@
+# Custom Service
+
+An example of using Gotham within a pre-existing Hyper service. This is useful
+for advanced use cases where you might want to wrap Gotham in order logic, or
+pre-populate your state handing it off to Gotham.
+
+## License
+
+Licensed under your option of:
+
+* [MIT License](../../LICENSE-MIT)
+* [Apache License, Version 2.0](../../LICENSE-APACHE)
+
+## Community
+
+The following policies guide participation in our project and our community:
+
+* [Code of conduct](../../CODE_OF_CONDUCT.md)
+* [Contributing](../../CONTRIBUTING.md)
--- a/examples/custom_service/src/main.rs
+++ b/examples/custom_service/src/main.rs
@@ -1,0 +1,78 @@
+//! An example usage of Gotham from another service.
+
+use anyhow::{Context as _, Error};
+use futures::future::{BoxFuture, FutureExt};
+use gotham::{
+    router::{builder::*, Router},
+    service::call_handler,
+    state::State,
+};
+use http::{Request, Response};
+use hyper::{server::conn::Http, service::Service, Body};
+use std::net::SocketAddr;
+use std::panic::AssertUnwindSafe;
+use std::task;
+use tokio::net::TcpListener;
+
+#[derive(Clone)]
+struct MyService {
+    router: Router,
+    addr: SocketAddr,
+}
+
+impl Service<Request<Body>> for MyService {
+    type Response = Response<Body>;
+    type Error = Error;
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> task::Poll<Result<(), Self::Error>> {
+        task::Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        // NOTE: You don't *have* to use call_handler for this (you could use `router.handle`), but
+        // call_handler will catch panics and return en error response.
+        let state = State::from_request(req, self.addr);
+        call_handler(self.router.clone(), AssertUnwindSafe(state)).boxed()
+    }
+}
+
+pub fn say_hello(state: State) -> (State, &'static str) {
+    (state, "hello world")
+}
+
+#[tokio::main]
+pub async fn main() -> Result<(), Error> {
+    let router = build_simple_router(|route| {
+        // For the path "/" invoke the handler "say_hello"
+        route.get("/").to(say_hello);
+    });
+
+    let addr = "127.0.0.1:7878";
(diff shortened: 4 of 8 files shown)


krallin:
The clippy errors that are failing in CI look to be pre-existing here

msrd0:
Thanks for the PR!

I'm not sure I like everyone being able to create instances of `State` though ... so far, the only ways to obtain a `State` is either through a handler or through `State::with_new`. The first one gives you an owned version but requests it back as the return type of the handler, the second one only passes your callback a mutable reference. As long as we don't force all handlers to take mutable references, which will have to wait at least until async closures are properly supported on stable rust, I don't think I want any way to obtain an owned version of the State that is not made inaccessible through `pub(crate)` or made temporary as in you have to return its ownership back to gotham.

If I understand your use case correctly, you want to call a gotham handler from a raw hyper response and not use gotham to interface with hyper. This should be possible if you changed your handlers to take mutable references to the state, and converted the `from_request` method to one that takes an already existing state and only populates it with the information from the request. This way, gotham could internally call `let mut state = State::new(); state.populate_from_request(request);` and you could use `State::with_new` to call your own handlers. (Feel free to come up with a better name than `populate_from_request`)

And don't worry about the clippy errors, that check is marked optional but GitHub doesn't really display that anywhere.

krallin:
> If I understand your use case correctly, you want to call a gotham handler from a raw hyper response and not use gotham to interface with hyper

Not quite — let me rephrase. I have a `hyper::Request` + some extra context I extracted from the client's socket that I want to throw into the state. Then, I just want to call this method on my existing Gotham `Handler`: 

https://github.com/gotham-rs/gotham/blob/master/gotham/src/handler/mod.rs#L167

That method doesn't really take a `&mut State`, but even if it it did, I don't think I could use `State::with_new`. If Gotham gives me a borrowed state, my return value's lifetime can't exceed the lifetime of the borrowed state. Unfortunately, my return value is a future that needs to be `'static`.

Here's what my code looks like. This might help:

```rust
impl<H: Handler + Clone + Send + Sync + 'static + RefUnwindSafe> MononokeHttpHandlerAsService<H> {
    pub async fn call_gotham(self, req: Request<Body>) -> Response<Body> {
        let mut state = State::from_request(req, self.addr);
        if let Some(tls_socket_data) = self.tls_socket_data {
            tls_socket_data.populate_state(&mut state);
        }

        let res = match self.handler.handle(state).await {
            Ok((_state, res)) => res,
            Err((state, err)) => err.into_response(&state),
        };

        res
    }
}

// This just a little bridge to Hyper for places where we want to use this directly in `serve_connection`.
impl<H: Handler + Clone + Send + Sync + 'static + RefUnwindSafe> Service<Request<Body>>
    for MononokeHttpHandlerAsService<H>
{
    type Response = Response<Body>;
    type Error = anyhow::Error;
    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> task::Poll<Result<(), Self::Error>> {
        task::Poll::Ready(Ok(()))
    }

    fn call(&mut self, req: Request<Body>) -> Self::Future {
        let this = self.clone();
       
        async move {
            let res = this.call_gotham(req).await;
            Ok(res)
        }
        .boxed()
    }
}
```

Can you share a little more about the reasons behind the objection to have an owned state exposed to the client? 

Thanks!

msrd0:
Ok I think I understand what you're trying to acchieve - still I think my suggestion would work, you'd just have to define your own trait for a handler that takes `&mut State`, similar to [`AsyncHandlerFn`](https://github.com/gotham-rs/gotham/blob/master/gotham/src/router/builder/single.rs#L26).

As to not allowing users to create an owned version of `State` - this has been impossible before I've gotten involved in gotham, so I don't now the original reasoning behind `State::new` being private. I'd assume the reason is to ensure that all user-supplied handlers return the state they were passed initially, and don't create a new state to return from the handler, destroying the state they were passed as argument. At least I think this is a good reason to disallow the creation of owned states outside of gotham's internal code.

krallin:
> Ok I think I understand what you're trying to acchieve - still I think my suggestion would work, you'd just have to define your own trait for a handler that takes `&mut State`, similar to `AsyncHandlerFn`.

But then I can't use a Gotham Router as my handler ... so I'm not using Gotham anymore?

msrd0:
Oh yeah you're right that won't work ... what you're trying seems related to #183, right? I have never had this problem before so I'm not sure what the best way to solve this is. Having `State::new` private and `State::from_request` public seems weird at first, but might make sense since it makes it a little bit harder to return a newly created state from a handler instead of the original one.

I'm just trying to keep the api straight-forward, and returning `(State::new(), create_response(...))` seems straight-forward but is incorrect, so I want to make sure nobody thinks something like this would be correct.

krallin:
> Oh yeah you're right that won't work ... what you're trying seems related to #183, right? I have never had this problem before so I'm not sure what the best way to solve this is

Yes, pretty much. I think one alternative here is to have Gotham pass through Extensions and make the "connect" mechanism a public interface, then I can take inject the extra state I have from Extensions into the Gotham state early in my middleware stack.

Alternatively, Gotham could have 2 hyper::Service implementations. The "connected" one can just inject the client address into extensions (and stay private), and the "not connected" one can just take them from Extensions, then callers can then add their own client address into extensions. Gotham might have to tolerate the client address being missing though.

Thoughts? Would you be more amenable to that kind of patch?

msrd0:
No I think exposing `State::from_request` might be the best way to support this, and letting users create their own `Service` implementation if the one gotham provides does not fit their needs. However, this means that there are no panic handlers by default for custom services, I'm not sure if this is desirable. Do you think it would make sense to make `gotham::service::trap::call_handler` public? You could then use it instead of `self.handler.handle(state)` in your `call_gotham` method.

Regarding the code, could you add a small example (similar to the one you gave above) that demonstrates how to use gotham with a custom service? Also, please move the `from_request` method to the top below the `with_new` method, I'd prefer factory methods to appear before instance methods in the `impl` block.

krallin:
> Do you think it would make sense to make gotham::service::trap::call_handler public? You could then use it instead of self.handler.handle(state) in your call_gotham method.

I have no objections at all, though at least as far as I'm concerned I'm not really planning to use it (we configure all our services to abort on panics). I'm happy to fold this into this change.

> Regarding the code, [...]

Yeah, no problem, I'll update the PR with this. Thanks!

krallin:
I updated the PR with an example and moved the code around. Let me know if I should also make the trap handler public and perhaps use it in the example. Thanks! :)

msrd0:
> Let me know if I should also make the trap handler public and perhaps use it in the example.

Yes, please do. This should make the example service behave exactly as the gotham provided service, thus should be a good starting point for everyone that needs to customize their service.

krallin:
> Yes, please do. This should make the example service behave exactly as the gotham provided service, thus should be a good starting point for everyone that needs to customize their service.

Cool; I updated the PR to use `call_handler` :)

msrd0 (approved):
Thanks

krallin:
@msrd0 :do you think we could get a release with those changes at some point? I'd love to get us off of our fork :)

msrd0:
I'm currently waiting on #468, after that I'll get a release out.

msrd0:
FYI: I just released version 0.6.0

PR #534, opened 2021-04-05
Fix clippy lints and make clippy check required

msrd0:
- Fixed all clippy lints currently in the code
- Formated all examples in doc comments (using a nightly version of rustfmt)
- Make the clippy check required in GitHub Actions

I opted to not fix the `upper_case_acronyms` lint as renaming `JWTMiddleware` to `JwtMiddleware` would be a breaking change.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -48,7 +48,6 @@ jobs:
   
   clippy:
     runs-on: ubuntu-latest
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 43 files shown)


nyarly (approved):
All looks reasonable.

Should we begin the process (in a new PR) of aliasing and deprecating the JWT constant?

msrd0:
> Should we begin the process (in a new PR) of aliasing and deprecating the JWT constant?

@nyarly Sounds good, I can create a PR if you want

PR #552, opened 2021-08-15
Overhaul TestServer in preparation of AsyncTestServer

FSMaxB:
While implementing an `AsyncTestServer` I did a lot of changes to the existing `TestServer` as well especially in order to reuse existing code.

I also noticed that `client_with_address` was broken since gotham 0.3.0, and fixing it isn't really possible without copy-pasting most of the `bind_server` function into the `TestServer` implementation. So in this pull request I'm deprecating it and removing misleading documentation that has been wrong for several releases now. If requested, I can still fix the `client_with_address` feature, though at the cost of copy-pasted code as mentioned.

Things I did roughly:
* Pull out the `TestHandler` into a common module and increase the ergonomics of it's usage.
* Fix the `timeout` tests.
* Change the `bind_server_rustls` to a reusable `rustls_wrap` function instead. This allows using the same `bind_server` code for both `tls` and `plain` `TestServer`
* Move most `TestServer` code into the internal `TestServerData` so it can be reused by both of them.
* Deduplicated the `TestServer` tests so both the `plain` and `tls` tests reuse the same test code.


FSMaxB:
Don't ask me what problem the coverage test is having again ...

msrd0:
Thanks for working on this!

> I also noticed that `client_with_address` was broken since gotham 0.3.0

I couldn't find any issues complaining about this, so the function seems to be completely unused. We should probably remove it altogether.

> Don't ask me what problem the coverage test is having again

Some (unused, but un-removable) dependency of askama doesn't compile if some tarpaulin-specific cfg is activated. For the time being, #554 should fix this. Once askama uses (yet-unreleased) nom 7.0, this workaround should no longer be necessary.

msrd0:
codecov is working again on the master branch. Can you merge/rebase please?

FSMaxB:
> I couldn't find any issues complaining about this, so the function seems to be completely unused. We should probably remove it altogether.

I will leave the deprecation notice in for now. My intention was not to break the API for now (if I accidentally did, please say so!). It can still be removed in a separate pull request.

> codecov is working again on the master branch. Can you merge/rebase please?

Done. Thank's for fixing.

FSMaxB:
I'm going to at least make some limited efforts to try to get codecov happy.

msrd0 (approved):
Thanks

PR #553, opened 2021-08-15
AsyncTestServer

### LINKED ISSUE #448: TestRequest should have perform async method
Hi,
I have a situation when there are two clients. One pushes some updates to the server, and the other one polls. 
I want to be able to test this concurrent scenario. Currently, all test requests are done with `perform()` method which is blocking.
However, the internals of the TestRequest use futures, but I don't see any way to access them from public api. 

Do you think we can have something like `perform_async()` method, that will return a future? Afterward all futures can be awaited before exiting test method.

FSMaxB:
Fixes #448 

NOTE: This is based on #552, so it contains all the commits that the other pull request contains. If changes are made to #552, this PR will be rebased on top of that.

This finally implements a test server that doesn't have it's own runtime and can therefore be run on any tokio runtime using `async` and `await` without the need of awkward hacks with `run_future`.

Note that it would be completely possible to not have `AsyncTestClient`, `AsyncTestRequestBuilder` and `AsyncTestResponse` and just use `hyper::Client` directly here, but the ergonomics of a raw `hyper::Client` isn't that great, that's why I implemented these helpers. The `AsyncTestRequestBuilder` implements `Into<hyper::Client>` though, so the option to use that directly is still there.

At least one open question I'm asking myself: Should the `AsyncTestClient` have a clone of the `Arc<AsyncTestServerInner>` so that the server is kept alive as long as there is still a remaining client? Otherwise requests via the `AsyncTestClient` will fail if the `AsyncTestServer` has been dropped!


FSMaxB:
Also note that I've added some features to the `tokio` dependency of gotham. If that is undisired, a separate tokio dependency could be added to the `dev-dependencies` and the `resolver = "2"` be added to the `Cargo.toml` so that the `dev-dependencies` features don't end up in the regular dependency.

FSMaxB:
I'm going to rebase this once again since #552 was merged!

msrd0:
Thanks for implementing this! The `src/{plain,tls}/async_test.rs` files are rather small, I think they could just be part of the existing `test.rs` files? Gotham already has a lot of (user-exposed) modules.

> Also note that I've added some features to the tokio dependency of gotham. If that is undisired, a separate tokio dependency could be added to the dev-dependencies and the resolver = "2" be added to the Cargo.toml so that the dev-dependencies features don't end up in the regular dependency.

Yes, please make them dev-dependencies if they aren't needed to compile non-test code.

FSMaxB:
> I think they could just be part of the existing test.rs files? Gotham already has a lot of (user-exposed) modules.

> Yes, please make them dev-dependencies if they aren't needed to compile non-test code.

Will do both of those. Although `resolver = "2"` doesn't seem to be an option. But this was based on a misunderstanding on my part anyways and isn't necessary!

Any feedback on keeping a clone of `Arc<AsyncTestServerInner>` in the `AsyncTestClient` or not?

msrd0:
> Any feedback on keeping a clone of `Arc<AsyncTestServerInner>` in the `AsyncTestClient` or not?

Looking at the current sync version, which does store a clone of the server, we should probably do the same for the async version, just for consistency. I don't really like stuff not getting dropped when it goes out of scope, but it seems like the clone is actually required for the sync version, so there isn't really much we can do about it.

FSMaxB:
Should `gotham::async_test` module be kept `pub(crate)` and the `AsyncTestClient`, `AsyncTestRequestBuilder` and `AsyncTestResponse` be exported as `pub` from `gotham::test`?

msrd0:
> Should `gotham::async_test` module be kept `pub(crate)`

Yes

> AsyncTestClient, AsyncTestRequestBuilder and AsyncTestResponse be exported as pub from gotham::test

Wouldn't those be exported from `gotham::plain::test` and `gotham::tls::test`?

FSMaxB:
> Wouldn't those be exported from gotham::plain::test and gotham::tls::test?

No they are independent of the concrete `AsyncTestServer` implementation. They are generic over `C: Connect` and that's it. Same as with the `TestClient` and `TestResponse` types that are in `gotham::test`

msrd0 (approved):
Thanks

PR #577, opened 2021-11-10
Add example of middleware with async usage

AngelicosPhosphoros:
I think, this is useful because allows user to avoid usage of futures_util.

I don't know if I must adjust existing introduction or add new example project to the middleware folder so I added it to existing example for now.

I can change it if needed.


msrd0:
Hi, thanks for the PR!

I think it makes more sense to add a second example (say introduction_await) that does the same as the current example, just with using await instead of futures-util's function calls.

AngelicosPhosphoros:
I added new example project and removed changes from old example.

msrd0 (approved):
Thanks

PR #578, opened 2021-11-10
Replace chrono dependency by time

AngelicosPhosphoros:
[chrono](https://crates.io/crates/chrono) has not been released for very long time (1 year already).
Also, it has security error: https://rustsec.org/advisories/RUSTSEC-2020-0159.html

Also, this would reduce dependencies because `chrono` uses very old `time` version so many binaries which use gotham ends with 2 versions of `time` linked.


AngelicosPhosphoros:
There is new contributor to chrono so they could fix chrono in future.
https://github.com/chronotope/chrono/issues/606#issuecomment-951326791

AngelicosPhosphoros:
Also, this change reduce complexity of `Timing` type because there is no more `Invalid` state.

AngelicosPhosphoros:
Also, I suggest to add `std::time::Instant` field to `Timer` struct to get more valid durations.
Old and current solutions can lead to negative durations which is not very good.

So we would use `time` to log start time but we would use `Instant` to get duration of requests.

msrd0:
> Also, I suggest to add std::time::Instant field to Timer struct to get more valid durations.

I was just thinking exactly that. That also removes the Invalid case as it guarantees monotonicity.

msrd0 (approved):
Thanks

PR #601, opened 2022-09-02
Update diesel to 2.0

msrd0:


nyarly (approved):
:+1: We'll need to consider semver on this

msrd0:
> We'll need to consider semver on this

diesel is a public dependency of the diesel middleware so this is definitely a breaking change

svenfoo:
Is there a plan to get this merged? Just finished porting the ORM parts of a project to diesel-2.0 only to find out that ` gotham_middleware_diesel` isn't compatible with it yet.  Would be awesome to get this merged and released soon.

msrd0:
@svenfoo I think my plan was to migrate my server code to diesel 2.0 and test these changes before I merge and release them, but that never happened. Given that the changes seems rather simple and we have tests, it's probably ok to merge this anyways.

PR #623, opened 2023-09-12
Bump minimal versions of anyhow, thiserror and serde

msrd0:
We actually require `anyhow` 1.0.5 and upwards to allow `anyhow::Error` to be used with hyper, and `thiserror` 1.0.2 to use the `#[from]` attribute. I also bumped serde to the oldest version after switching back to from-source compilation.


msrd0:
For some reason, when using `-Z minimal-versions` the diesel example's test fails. But at least everything compiles, so this is still an improvement

PR #628, opened 2023-10-11
assets: add range request support for FileHandler

jpelkonen:
This adds range request support for FileHandler. This is a critical feature for handling large files, or any application that will require range request support, e.g. video or audio streaming.

This implementation currentyl only supports ranges that result in a single response, i.e. multiple ranges are not supported.

This implementation will return an error if a range is open from both ends, or if the range beginning is larger than the end.

### DIFF
--- a/gotham/src/handler/assets/mod.rs
+++ b/gotham/src/handler/assets/mod.rs
@@ -17,7 +17,7 @@ use mime::{self, Mime};
 use mime_guess::from_path;
 use serde::Deserialize;
 use tokio::fs::File;
-use tokio::io::{AsyncRead, ReadBuf};
+use tokio::io::{AsyncRead, AsyncSeekExt, ReadBuf};
 
 use self::accepted_encoding::accepted_encodings;
 use crate::handler::{Handler, HandlerError, HandlerFuture, NewHandler};
@@ -26,6 +26,7 @@ use crate::state::{FromState, State, StateData};
 
 use std::convert::From;
 use std::fs::Metadata;
+use std::io::{ErrorKind, SeekFrom};
 use std::iter::FromIterator;
 use std::mem::MaybeUninit;
 use std::path::{Component, Path, PathBuf};
@@ -206,7 +207,7 @@ fn create_file_response(options: FileOptions, state: State) -> Pin<Box<HandlerFu
 
     let (path, encoding) = check_compressed_options(&options, &headers);
 
-    let response_future = File::open(path).and_then(|file| async move {
+    let response_future = File::open(path).and_then(move |mut file| async move {
         let meta = file.metadata().await?;
         if not_modified(&meta, &headers) {
             return Ok(hyper::Response::builder()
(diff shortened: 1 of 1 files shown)


gotham/src/handler/assets/mod.rs
```diff
@@ -206,16 +207,26 @@ fn create_file_response(options: FileOptions, state: State) -> Pin<Box<HandlerFu
...
-        let len = meta.len();
         let buf_size = optimal_buf_size(&meta);
+        let resolve_range_result = resolve_range(meta.len(), &headers);
+        if let Err(e) = resolve_range_result {
+            return Ok(hyper::Response::builder()
+                .status(StatusCode::RANGE_NOT_SATISFIABLE)
+                .body(Body::from(e))
+                .unwrap());
+        }
+        let (len, range_start) = resolve_range_result.unwrap();
```
msrd0, line 226:
```suggestion
        let (len, range_start) = match resolve_range(meta.len(), &headers) {
            Ok((len, range_start)) => (len, range_start),
            Err(e) => {
                return Ok(hyper::Response::builder()
                    .status(StatusCode::RANGE_NOT_SATISFIABLE)
                    .body(Body::from(e))
                    .unwrap());
            }
        };
```

No need to use `.unwrap()` if the compiler can prove it.

gotham/src/handler/assets/mod.rs
```diff
@@ -251,6 +275,52 @@ fn create_file_response(options: FileOptions, state: State) -> Pin<Box<HandlerFu
         .boxed()
 }
 
+// Checks for existence of "Range" header and whether it is in supported format
+// This implementations only supports single part ranges.
+// Returns a result of length and optional starting position, or an error if range value is invalid
+// If range header does not exist or is unsupported the length is the whole file length and start position is none.
```
msrd0, line 281:
```suggestion
/// Checks for existence of "Range" header and whether it is in supported format
/// This implementations only supports single part ranges.
/// Returns a result of length and optional starting position, or an error if range value is invalid
/// If range header does not exist or is unsupported the length is the whole file length and start position is none.
```

Please use doc comments to document functions.

gotham/src/handler/assets/mod.rs
```diff
@@ -251,6 +275,52 @@ fn create_file_response(options: FileOptions, state: State) -> Pin<Box<HandlerFu
...
+                            (None, Some(end)) => {
+                                let begin = len.saturating_sub(end);
+                                Ok((end, Some(begin)))
+                            }
+                            (None, None) => Err("invalid range"),
+                        }
+                    })
+            })
+        })
+        .unwrap_or(Ok((len, None)))
```
msrd0, line 321: This code block looks way too complex. Try begining with

```rust
let range = headers.get(RANGE) else { return Ok((len, None)); };
```

gotham/src/handler/assets/mod.rs
```diff
@@ -841,6 +913,62 @@ mod tests {
...
+                .client()
+                .get(format!("http://localhost/{file_name}"))
+                .with_header(RANGE, HeaderValue::from_str(&range_header).unwrap())
+                .perform()
+                .unwrap();
+            if range_start == 0 && range_len == 0 {
+                assert_eq!(response.status(), StatusCode::RANGE_NOT_SATISFIABLE);
+                break;
+            }
+            std::io::Seek::seek(&mut file, SeekFrom::Start(range_start)).unwrap();
```
msrd0, line 949: This looks odd. Try importing `std::io::Seek` and then calling `file.seek(...)`.

msrd0 (changes requested):
Thanks for your PR! I've added a few comments, the rest looks really good

gotham/src/handler/assets/mod.rs
jpelkonen: I have changed this implementation

msrd0: GitHub shows no change. Did you forget to push?

jpelkonen: You're right, my mistake

msrd0: Thanks

msrd0 (approved):

Issue #631, opened 2023-10-25
Range request response status code should be 206 rathen than 200


jpelkonen:
From https://www.rfc-editor.org/rfc/rfc9110.html#section-14.2:

> If all of the preconditions are true, the server supports the Range header field for the target resource, the received Range field-value contains a valid [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) with a [range-unit](https://www.rfc-editor.org/rfc/rfc9110.html#range.units) supported for that target resource, and that [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) is [satisfiable](https://www.rfc-editor.org/rfc/rfc9110.html#satisfiable) with respect to the selected representation, the server SHOULD send a [206 (Partial Content)](https://www.rfc-editor.org/rfc/rfc9110.html#status.206) response with content containing one or more partial representations that correspond to the satisfiable [range-spec](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier)(s) requested.

The current implementation sends 200.

PR #633, opened 2023-10-25
assets: Fix range request status code

jpelkonen:
Should return partial content (206) rather than (200) RFC 9110, section 14.2: "the server SHOULD send a 206 (Partial Content) response with content containing one or more partial representations"

This commit addresses issue #631


msrd0 (approved):
Thanks

PR #645, opened 2024-02-28
Fix feature = "cargo-clippy" deprecation

rex4539:
https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html


msrd0 (approved):
Thanks

msrd0:
Hm I guess older clippy versions don't recognize `cfg(clippy)` which makes sense. Do we need the `cfg_attr(clippy)` anyways? The blog post seems to read like they could just be removed these days

rex4539:
Applied the correct fix.

msrd0 (approved):
Thanks

